Issue #632, opened 2014-07-03
Accumulator sum


suman724:
I am using Morphia 0.108

I am trying to use the static method sum in the class org.mongodb.morphia.aggregation.Group. This method takes a string as a parameter and is returning an Accumulator. The parameter I want to pass is a constant instead of key.

For example - I am using this in an aggregation query to count customers by priority. 

{ $group : { "_id" : "$priority", "customersPerStatus" : { $sum : 1 } }} 

I did not find a way in Morphia framework to pass constants to the Sum accumulator. 

Any comments?

evanchooly:
fixed with the merge of #666

PR #666, opened 2014-10-31
Allow Proper Implementation of Accumulator Sum

AddoSolutions:
See Issue #632

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/aggregation/Accumulator.java
+++ b/morphia/src/main/java/org/mongodb/morphia/aggregation/Accumulator.java
@@ -2,14 +2,18 @@ package org.mongodb.morphia.aggregation;
 
 public class Accumulator {
     private final String operation;
-    private final String field;
+    private final Object field;
 
     public Accumulator(final String operation, final String field) {
+        this(operation, (Object)("$" + field));
+    }
+	
+    public Accumulator(final String operation, final Object field) {
         this.operation = operation;
-        this.field = "$" + field;
+        this.field = field;
     }
 
-    public String getField() {
+    public Object getField() {
         return field;
     }
 


evanchooly:
Given that `field` is a name, i'm not sure what changing this to `Object` really accomplishes.  Given the time lapsed since a test was requested with no response, I'll probably end up just closing this one unless I hear something soon.

AddoSolutions:
You cannot give an integer for `$sum`, for example, to tell morphia to sum the values.

exmaple:

```
		AggregationPipeline<Stat, AggregateTarget> pipeline
			= getDs().<Stat, AggregateTarget>createAggregation(Stat.class)
			.match(query)
			.group("target", grouping("total", new Accumulator("$sum", 1)))
			.sort(new Sort("total", -1))
			.limit(limit);
```

Notice on the third line: `new Accumulator("$sum", 1)` – if that is a String, it is parsed as `{ $sum : "1" }` when in order for Mongo to do the indexing, it must be: `{ $sum : 1 }` as an integer or in some cases a BasicDBObject.

See: http://docs.mongodb.org/manual/reference/operator/aggregation/sum/

evanchooly:
oh!  right.  I see.  That makes way more sense now.  Thanks.

evanchooly:
@AddoSolutions can you add a unit test or two to the pull request? I can merge it for 0.111.

AddoSolutions:
Not to n00b you out, but could you give me an example of what you are looking for?  Thanks :)

evanchooly:
e.g., when I added the new text indexing bits, i added this test:  https://github.com/mongodb/morphia/blob/master/morphia/src/test/java/org/mongodb/morphia/indexes/TestTextIndexing.java

Something that shows it in use for the users and something that shows it works for the devs.

PR #695, opened 2015-01-12
Downgrade from WARN to DEBUG message about capped collection existing (fixes #694)

### LINKED ISSUE #694: Erroneous warning "DatastoreImpl: DBCollection already exists is capped already"
When Morphia manages an entity stored in a capped collection, the first time the application starts, Morphia creates the capped collection.  On the second and all subsequent starts of the application, Morphia logs a warning like this:

```
WARN  [2015-01-12 20:57:35,024] org.mongodb.morphia.DatastoreImpl: DBCollection already exists is capped already; doing nothing. { "serverUsed" : "localhost:27017" , ...}
```

Looking at the logic in [DataStoreImpl](https://github.com/mongodb/morphia/blob/f765a7095cbcd1bf46a94f7047db88b17d2f60ce/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java#L438), it's clear that this log message is printed every time `ensureCaps()` is called.  This log message should probably be DEBUG level or removed entirely.

jbyler:
Sample warning being changed to debug level:

```
  WARN org.mongodb.morphia.DatastoreImpl: DBCollection already exists is capped already; doing nothing.
```

This should not be a warning as there is no problem.  The DBCollection
will already exist on every application startup except the first one.
Also fixes grammar of the log message.

Fixes Morphia issue #694.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
@@ -435,7 +435,7 @@ public class DatastoreImpl implements AdvancedDatastore {
                     final DBObject dbResult = database.command(BasicDBObjectBuilder.start("collstats", collName).get());
                     if (dbResult.containsField("capped")) {
                         // TODO: check the cap options.
-                        LOG.warning("DBCollection already exists is capped already; doing nothing. " + dbResult);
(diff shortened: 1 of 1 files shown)


jbyler:
Perfect, thanks much!

PR #696, opened 2015-01-13
improve performance of multimaps

scottbessler:
Added a Perf test and ~4x improved the performance for mapping of multimaps (```Map<String,List<Value>>```)

before:
```
Fetch multimap performance is too slow: 4.03X slower (10375/2575)
```

after: 
```~1.38x slower than raw```

further, added a fix to allow jdk8 to build the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,3 @@ target
 bin
 .gradle
 out/
-**/gradle.properties
--- a/build.gradle
+++ b/build.gradle
@@ -49,7 +49,7 @@ allprojects {
     }
 
     group = 'org.mongodb.morphia'
-    version = '0.110-SNAPSHOT'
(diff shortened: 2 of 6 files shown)


scottbessler:
i also re-included gradle.properties as that is best practice and useful for auto-versioning with things like teamcity. I apologize for the mish-mash of changes, but I had a hard time developing because of some of the gradle issues.

scottbessler:
thanks!

evanchooly:
no problem

*--------------------------------*

*{ *name     : *"**Justin Lee**", *
  title    : *"*Software Engineer*"*,
  twitter  : *"*@evanchooly <http://twitter.com/evanchooly>*"*,
  web      : *[* "10gen.com", *"*antwerkz.com*"* *]*,
  location : *"*New York, NY*"* *}*

Issue #643, opened 2014-08-01
Create types for all supported GeoJson types


trishagee:
The current geospatial support in Morphia assumes entities storing location data as [legacy coordinate pairs](http://docs.mongodb.org/manual/reference/glossary/#term-legacy-coordinate-pairs).  Since version 2.4 of the server, the preferred approach for storing location information is [GeoJson](http://docs.mongodb.org/manual/applications/geospatial-indexes/#geo-overview-location-data).

Morphia needs to support entities that store location as GeoJson.  My proposal is to create types for each of the GeoJson objects, so users can simply create variables of these types in their entities if they wish to store GeoJson data in their entities.

We should support all currently supported GeoJson types:
Version 1 (for server versions 2.4 and higher):
 - Point
 - LineString
 - Polygon

Version 2 (for server versions 2.6 and higher):
 - MultiPoint
 - MultiLineString 
 - MultiPolygon
 - GeometryCollection

Saving an entity with one of these types in it will result in the creation of a document containing valid GeoJson.

This is part of the work to complete #589, and is required in order to complete #642, #644, and #645

PR #700, opened 2015-01-22
API for $near queries.  Fixes #642

### LINKED ISSUE #642: Support $near queries on entities containing GeoJson
Currently the `near` queries in Morphia assume entities storing location data as [legacy coordinate pairs](http://docs.mongodb.org/manual/reference/glossary/#term-legacy-coordinate-pairs) and create [$near](http://docs.mongodb.org/manual/reference/operator/query/near/) queries that work off this assumption.

Update Morphia to support geospatial "near" queries that work with entities containing locations stored as [GeoJson](http://docs.mongodb.org/manual/applications/geospatial-indexes/#geo-overview-location-data).  

This should support v1 GeoJson (for server versions 2.4 and higher):
 - Point
 - LineString
 - Polygon

And v2 GeoJson (for server versions 2.6 and higher):
 - MultiPoint
 - MultiLineString 
 - MultiPolygon
 - GeometryCollection

This is part of the work to complete #589

trishagee:
Based on the work done for #643, added new `near` methods that use the new Point objects and take a distance parameter (instead of radians).

Note that the entity types need a different converter when they're passed in as a query parameter than when they're in an actual entity, as the underlying MongoDB BSON is different.

Note that most of this code is test code.


evanchooly:
merged 119c08d28c6ff267306cbe8d09d835ab03d037e6

PR #701, opened 2015-01-24
Add option to lowercase default entity collection name

RainoBoy97:
Option to lowercase the default entity collection name when nothing is specified in the @Entity annotation!

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/mapping/MappedClass.java
+++ b/morphia/src/main/java/org/mongodb/morphia/mapping/MappedClass.java
@@ -538,7 +538,10 @@ public class MappedClass {
      * @return the collName
      */
     public String getCollectionName() {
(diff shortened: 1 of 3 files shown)


ParaPenguin:
Would also be nice if capital letters in the class name were prepended with an underscore.

evanchooly:
This is great.  But it'd be event better with tests.  :)

RainoBoy97:
Added a test. I used the already existing `MapperOptionsTest` class because it seemed like an appropriate place to put it.

The commit messages got messed up on my fork, but thankfully it didnt mess up the pull request :)

RainoBoy97:
Since you added it manually and not merging I guess this can be closed.

evanchooly:
It is closed.

RainoBoy97:
Oh wow, how did I not notice that...

PR #725, opened 2015-03-16
Text index approaches

evanchooly:
implementing text indexing in two different fashions.  let the voting commence.


morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -384,6 +390,102 @@ private void processClassAnnotations(final DBCollection dbColl, final MappedClas
                 }
             }
         }
+        
+        mapTextIndexed(dbColl, mc);
```
evanchooly: These two method calls are the respective hooks for each approach

morphia/src/test/java/org/mongodb/morphia/indexes/TestTextIndexing.java
```diff
@@ -0,0 +1,91 @@
...
+import org.mongodb.morphia.annotations.Id;
+import org.mongodb.morphia.annotations.Language;
+import org.mongodb.morphia.annotations.Text;
+import org.mongodb.morphia.annotations.TextIndex;
+import org.mongodb.morphia.annotations.TextIndexed;
+
+import java.util.List;
+
+public class TestTextIndexing extends TestBase {
+    @Entity
```
evanchooly: First approach:  multiple annotations

morphia/src/test/java/org/mongodb/morphia/indexes/TestTextIndexing.java
```diff
@@ -0,0 +1,91 @@
...
+        @Text
+        private String name;
+        @Text(10)
+        private String nickName;
+        @Language
+        private String nativeTongue;
+    }
+
+    @Entity
+    @TextIndex(name = "singleAnnotation", language = "russian", value = {"name", "nickName"}, weights = {"nickName : 10"},
```
evanchooly: second approach one uber-annotation

evanchooly:
Here are two approaches to defining text indexes:  the first uses 3 annotations; the second only one.  I lean slightly toward the first approach as it's a bit easier to parse and verify the values.  The second on is definitely the one stop shop but requires a bit of duplication between the 'value' and 'weights' parameters and then there's that ugly string parsing needed to get at the weight values.  Since it's a string there's no way to enforce that only int values are passed in until validated at runtime.

Still, here are two options.  Which feels better?

trishagee:
The first is arguably uglier, but it does allow much greater flexibility for extension. And given we don't know what sort of indexes are coming in the future, it feels a bit safer.

Issue #230, opened 2013-04-25
Problem with using many DataStores [moved]


jyemin:
This is [Issue 230](http://code.google.com/p/morphia/issues/detail?id=230) moved from a Google Code project.
Added by 2011-02-24T07:54:40.000Z by [google-a...@codesmell.de](http://code.google.com/u/113306752357053708143/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Medium

### Original description

    
    package com.nexuus.jobs.data.repository;
    
    import java.net.UnknownHostException;
    
    import org.junit.Before;
    import org.junit.Test;
    
    import com.google.code.morphia.Datastore;
    import com.google.code.morphia.Morphia;
    import com.google.code.morphia.utils.Assert;
    import com.mongodb.DB;
    import com.mongodb.Mongo;
    import com.mongodb.MongoException;
    import com.nexuus.jobs.data.domain.Category;
    
    public class CategoryUweTest {
        private Category phoneCategory = new Category();
        private Category iphoneCategory = new Category();
        private Category appleCategory = new Category();
        private CategoryRepositoryDAO2 categoryDAO;
    
        protected Mongo mongo;
        protected Datastore ds;
        protected Morphia morphia = new Morphia();
    
        @Before
        public void setUp() {
            try {
                this.mongo = new Mongo();
            } catch (Exception e) {
               
            }
               
            this.ds = this.morphia.createDatastore(this.mongo, &quot;morphia_test&quot;);
            this.morphia.createDatastore(this.mongo, &quot;morphia_test2&quot;)
            this.categoryDAO = new CategoryRepositoryDAO2(ds);
        }
    
        @Test
        public void testCategoryTreeSavedAndRecovered() {
            phoneCategory.addChildrenCategory(appleCategory);
            appleCategory.setParentCategory(phoneCategory);
    
            appleCategory.addChildrenCategory(iphoneCategory);
            iphoneCategory.setParentCategory(appleCategory);
    
            categoryDAO.save(phoneCategory);
            categoryDAO.save(appleCategory);
            categoryDAO.save(iphoneCategory);
    
            phoneCategory = categoryDAO.get(phoneCategory.getId());
            appleCategory = categoryDAO.get(appleCategory.getId());
            iphoneCategory = categoryDAO.get(iphoneCategory.getId());
    
            Assert.isTrue(
                    phoneCategory.getChildrenCategories().contains(appleCategory),
[...]

PR #728, opened 2015-03-18
Allow provision of custom DatastoreProvider instances for new Morphia instances

kothar:
### Use case

I have a system which uses a single configuration datastore with one 'config' Morphia instance, and one datastore per customer, using an 'app' Morphia instance, to isolate their data and allow more granular control over data sets.

This presents a problem when interacting with the `DatastoreProvider`, since the default implementation uses a global singleton (`DatastoreHolder`) to record the most recently created `Datastore`.

When accessing the config store and the customer store at different times, the morphia instances will sometimes receive the wrong datastore for resolving `DBRef`s, thus causing errors.

### Changes
I have added a new constructor to `Mapper` which allows a custom `DatastoreProvider` to be passed in. I have also implemented two new `DatastoreProvider` strategies.

`SingleDatastoreProvider` will always return the datastore that was first created by the associated Morphia instance. This is on a per-`Mapper` basis, and prevents a mapper retrieving newer `Datastore`s which may have been created later.

`ThreadLocalDatastoreProvider` tracks the most recent datastore created on a per-thread basis rather than globally. This is also on a per-`Mapper` basis, which avoids datastores from other `Mapper` instances appearing unexpectedly.

This solves my particular use case, but there are obviously other strategies which could be used.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,7 +49,7 @@ allprojects {
     }
 
     group = 'org.mongodb.morphia'
-    // version = '0.110'
+    // version = '0.111-SNAPSHOT'
 
     sourceCompatibility = JavaVersion.VERSION_1_6
     targetCompatibility = JavaVersion.VERSION_1_6
--- a/build.gradle
+++ b/build.gradle
@@ -49,7 +49,6 @@ allprojects {
     }
 
     group = 'org.mongodb.morphia'
-    // version = '0.111-SNAPSHOT'
 
     sourceCompatibility = JavaVersion.VERSION_1_6
     targetCompatibility = JavaVersion.VERSION_1_6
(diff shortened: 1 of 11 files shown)


kothar:
This may also be of interest to issue #230 - a partial workaround may be to create a morphia instance for each database with separate mappers and a SingleDatastoreProvider.

Cross-db references would need something more complete however.

evanchooly:
When I pull in this PR, i get a lot of diffs against master which makes me think you might be slightly out of date.  Could you pull and rebase against master and repush?

kothar:
Sorry about that, yes, I was branched from the 0.110 tag.

ravikiran1226:
Still with the latest morphia i am having the problem with multiple databases

evanchooly:
Yes.  see #757

PR #729, opened 2015-03-18
Allow provision of custom DatastoreProvider instances for new Morphia instances

kothar:
This patch supersedes #728, and is branched from the current head.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
@@ -36,7 +36,6 @@ import org.mongodb.morphia.mapping.MappedField;
 import org.mongodb.morphia.mapping.Mapper;
 import org.mongodb.morphia.mapping.MappingException;
 import org.mongodb.morphia.mapping.cache.EntityCache;
-import org.mongodb.morphia.mapping.lazy.DatastoreHolder;
 import org.mongodb.morphia.mapping.lazy.proxy.ProxyHelper;
 import org.mongodb.morphia.query.DefaultQueryFactory;
 import org.mongodb.morphia.query.Query;
@@ -93,7 +92,7 @@ public class DatastoreImpl implements AdvancedDatastore {
         db = mongoClient.getDB(dbName);
 
         // VERY discussable
-        DatastoreHolder.getInstance().set(this);
+        mapper.getDatastoreProvider().register(this);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


evanchooly:
Thanks.  I'll take a look at this in a bit.

evanchooly:
The API looks a little wonky.  I noticed there are no tests.  Could you add some showing how this change would look when providing a custom provider?  The API feels a bit inverted and I'd like to see it in action.

mzapletal:
What about removing ``DatastoreProvider.register`` and setting the datastore for example via the constructor in a specific implementation?

evanchooly:
Yeah, the register bit looks weird to me.  I finally had chance to pull this PR locally and i'll walk through it today, i hope.

evanchooly:
OK.  I understand what's going on with this a bit better now.  I don't like it but that's not this PR's fault as this Provider thing is really half-backed.

So with that that thought, what exactly is this change supposed to provide.  The Provider doesn't really *do* much and I'm trying to decide what to do about that.  This PR touches upon an area that's seen a few requests for updates but internally morphia's just not structured to accommodate them just yet.  So if we're going to break the API to do this "right," now's the time before 1.0 locks us in to an API.

xeraa:
IMHO https://github.com/mongodb/morphia/pull/728#issue-62678389 describes it very well.

Our problem is the same: We have multiple datastores in a single application and get the wrong datastore when trying to fetch a $ref (which will obviously fail).

kothar:
I agree, this is probably the wrong way to approach the problem in the long term, this was just the minimal change that I could see: adding alternative implementations of the provider interface avoided changing the API for any existing code.

As @xeraa indicates, the change is supposed to provide a way to declare which Datastore instance is expected to be returned when the mapper is loading a DBRef, instead of the current behaviour of 'the last Datastore that was instantiated'. I would like to see some kind of distinct relationship between a Mapper and a Datastore, such that when working within a datastore instance there is no confusion over which location objects are being loaded from. This is the use case for the SingleDatastoreProvider class.

The problem with this is that the metadata collected by a Mapper could be used with several datastores containing the same model structure, and instantiating a new Mapper for each is slow and duplicates resources. Hence the ThreadLocalDatastoreProvider, which allows a single mapper to be used across multiple threads in a thread-per-request style application, and the correct datastore can then be set for each thread without duplicating Mapper instances per thread.

An alternative may be to share the metadata between mapper instances using some Mapper.clone() process applied during creation of the Datastore, thus allowing every datastore to be locked to a separate Mapper instance, which would then know which datastore to load references from. I'm not sure how this would interact with the dynamic mapping of new classes, but I think it would solve the same use case as described here.

evanchooly:
(Sorry for the delay.  I was out last week.)

I see the reasoning and I've figured out what I don't like about this pull request.  The good news is that it's not the PR to blame.  The bad news is that morphia's internal APIs are a bit of a mess.  I'll run the tests locally against this today and probably merge it since my concerns are no fault of yours.

evanchooly:
merged

PR #738, opened 2015-03-21
Made accumulator more abstract to be able to pass any value

LZaruba:
This way, users can create their own implementations of IAccumulator. Also existing API and contract of the current Accumulator implementation holds so users don't need to change their code.

Please note that I wasn't able to run the test suite so this change is "blind".

- extracted interface IAccumulator from Accumulator class to make it more abstract and to allow users to create their own implementationsmore abstract and to allow users to create their own implementations
- created GenericAccumulator to be able to pass generic value
- modified AggregationPipelineImpl accordingly

(issue #632)


evanchooly:
I really would like to see some tests on this.  Not only to validate the functionality but to give an example of how to use it and justify the change.

LZaruba:
Hi,
I've added a test to prove and show the functionality.
Please let me know if there is anything else needed.

Thanks

LZ

PR #762, opened 2015-04-30
provide a mechanism to override the coordinate system in geo queries

evanchooly:
pretty simple.  hopefully not too simple.


evanchooly:
@jyemin copied over that class if you want to take another look.  probably merge it later this afternoon.

PR #778, opened 2015-05-16
move annotations to separate project

### LINKED ISSUE #220: Put annotations in separate jar (no outside dependencies) [moved]
This is [Issue 220](http://code.google.com/p/morphia/issues/detail?id=220) moved from a Google Code project.
Added by 2011-02-02T04:48:06.000Z by [jasoncor...@gmail.com](http://code.google.com/u/106678326832894872424/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Low
### Original description

```
I'm creating a REST service for accessing my data, and I want to use the same interfaces on the server side as on the client side.  The problem is that the data classes that the rest service will return will have annotations on them for morphia.  So now I either need to create separate client interfaces, or have my client depend on morphia (and it's dependencies).  Neither of these is optimal.

You should provide a separate jar containing the annotations that has no dependencies and therefore can easily be included.

This isn't urgent, as I can create my own jar from your release, but it would be nice.
```

scottbessler:
this allows consumers to use morphia annotations in projects that
merely contain domain objects without forcing dependency on morphia and
transitively its dependencies.

this resolve #220 which was "gone away" for unclear reasons.


evanchooly:
Thanks for the pull request. Unfortunately, it's too late in the cycle to consider this for inclusion in 1.0. However, it begins to make more sense with certain future plans for Morphia and the Java driver. I've opened #782 to track this. By the time this enhancement is done, this pull request is extremely unlikely to merge cleanly so I'm going to mark this as closed and we can readdress it once other features are in place.

scottbessler:
are there not going to be any more releases beyond 1.0 ? seems like this could get on that next train.

scottbessler:
because waiting for a speculative feature request seems silly when this is already done.

PR #787, opened 2015-05-27
documentation for the Query API

evanchooly:
@jyemin @rozza

some updates to test to provide better examples or in some cases actually test features.

You can look over the code changes if you'd like.  The test class changes could probably use a little scrutiny.  The changes to Query.java are just a reorganization.

### DIFF
--- a/docs/reference/content/guides/querying.md
+++ b/docs/reference/content/guides/querying.md
@@ -75,7 +75,7 @@ The list of supported filter operations can be found in the [FilterOperator]({{<
 |$geoWithin | geoWithin |
 |$geoIntersects | geoIntersects |
 
- Each filter operator can either be referenced by its Mongodb "dollar operator" or by the aliases listed afterward.  For example, with 
+ Each filter operator can either be referenced by its MongoDB "dollar operator" or by the aliases listed afterward.  For example, with 
  the equal operator, you can use the canonical `$eq` operator as you would when building a query in the shell or you could opt to use 
  either the `=` or `==` aliases which might feel a little more natural to use than the dollar operators.
  
--- a/docs/reference/content/guides/querying.md
+++ b/docs/reference/content/guides/querying.md
@@ -119,10 +119,11 @@ method instead of `field()` but it is used in much the same fashion.  `and()` an
 ```java
 datastore.createQuery(UserLocation.class)
     .field("x").lessThan(5)
-    .field("y").greaterThan(4);
+    .field("y").greaterThan(4)
+    .field("z").greaterThan(10);
 ```
 
-This generates an implicit `and` between the two field comparisons.
+This generates an implicit `and` across the field comparisons.
 
 ## Text Searching
 
--- a/docs/reference/content/guides/querying.md
+++ b/docs/reference/content/guides/querying.md
@@ -113,8 +113,16 @@ q.and(
 
 An `or` clause looks exactly the same except for using `or()` instead of `and()`, of course.  For these clauses we use the `criteria()` 
 method instead of `field()` but it is used in much the same fashion.  `and()` and `or()` take a [`varargs`](https://docs.oracle
-.com/javase/8/docs/technotes/guides/language/varargs.html) parameter of type `Criteria` so you can include as many filters as necessary.
+.com/javase/8/docs/technotes/guides/language/varargs.html) parameter of type `Criteria` so you can include as many filters as necessary. 
+ If all you need is an `and` clause, you don't need an explicit call to `and()`:
+ 
+```java
+datastore.createQuery(UserLocation.class)
+    .field("x").lessThan(5)
+    .field("y").greaterThan(4);
+```
 
+This generates an implicit `and` between the two field comparisons.
 
 ## Text Searching
 
--- a/docs/reference/content/guides/querying.md
+++ b/docs/reference/content/guides/querying.md
@@ -45,10 +45,39 @@ In this case, we're instructing Morphia to add a filter using [$gte]({{< docsref
 ```
 
 The list of supported filter operations can be found in the [FilterOperator]({{< srcref
- "morphia/src/main/java/org/mongodb/morphia/query/FilterOperator.java">}}) class.  Each filter operator can either be referenced by its 
- Mongodb "dollar name" or by the aliases listed afterward.  For example, with the `EQUAL` operator you can use the canonical `$eq` operator
-  as you would when building a query in the shell, or you could opt to use either `=` or `==` which might feel a little more natural to 
-  use than the dollar operators.
+ "morphia/src/main/java/org/mongodb/morphia/query/FilterOperator.java">}}) class.  
+ 
+| Operator | Alias |
+| ---------|------- |
+|$center | |
+|$centerSphere | |
+|$box | |
+|$eq | =, == |
+|$ne | !=, <> |
+|$gt | > |
+|$gte | >= |
+|$lt | < |
+|$lte | <= |
+|$exists | exists |
+|$type | type |
+|$not | |
+|$mod | mod |
+|$size | size |
+|$in | in |
+|$nin | nin |
+|$all | all |
+|$elemMatch | elem, elemMatch |
+|$where | |
+|$near | near |
+|$nearSphere | |
+|$within (deprecated replaced by $geoWithin) | within |
+|$geoNear | geoNear |
+|$geoWithin | geoWithin |
+|$geoIntersects | geoIntersects |
+
+ Each filter operator can either be referenced by its Mongodb "dollar operator" or by the aliases listed afterward.  For example, with 
+ the equal operator, you can use the canonical `$eq` operator as you would when building a query in the shell or you could opt to use 
+ either the `=` or `==` aliases which might feel a little more natural to use than the dollar operators.
  
 ### `field()`
 
@@ -121,19 +150,11 @@ datastore.save(new Greeting("guten Morgen", "german"));
 datastore.save(new Greeting("guten Tag", "german"));
 datastore.save(new Greeting("guten Nacht", "german"));
 
-datastore.save(new Greeting("buenos días", "spanish"));
-datastore.save(new Greeting("buenas tardes", "spanish"));
-datastore.save(new Greeting("buenos noches", "spanish"));
-
 List<Greeting> good = datastore.createQuery(Greeting.class)
                              .search("good")
                              .order("_id")
                              .asList();
 Assert.assertEquals(4, good.size());
-Assert.assertEquals("good morning", good.get(0).value);
-Assert.assertEquals("good afternoon", good.get(1).value);
-Assert.assertEquals("good night", good.get(2).value);
-Assert.assertEquals("good riddance", good.get(3).value);
 ```
 
 As you can see here, we create `Greeting` objects for multiple languages.  In our test query, we're looking for occurrences of the work 
@@ -154,11 +175,15 @@ document.  This is useful when you need to only return a smaller view of a large
 ContainsRenamedFields user = new ContainsRenamedFields("Frank", "Zappa");
 getDs().save(user);
 
-ContainsRenamedFields found = getDs().find(ContainsRenamedFields.class).retrievedFields(true, "first_name").get();
+ContainsRenamedFields found = getDs()
+    .find(ContainsRenamedFields.class)
+    .retrievedFields(true, "first_name").get();
 Assert.assertNotNull(found.firstName);
 Assert.assertNull(found.lastName);
 
-found = getDs().find(ContainsRenamedFields.class).retrievedFields(true, "firstName").get();
+found = getDs()
+    .find(ContainsRenamedFields.class)
+    .retrievedFields(true, "firstName").get();
 Assert.assertNotNull(found.firstName);
 Assert.assertNull(found.lastName);
 ```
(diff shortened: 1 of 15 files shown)


docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+[menu.main]
+  parent = "Reference Guides"
+  weight = 86
+  pre = "<i class='fa fa-file-text-o'></i>"
++++
+
+# Querying
+
+Using the Java driver directly involves [creating](http://mongodb.github.io/mongo-java-driver/3.0/driver/reference/crud/#crud) a lot of 
+`Document` instances to capture a query.   Morphia, on the other hand, offers a fluent API with which to build up a query and attempts to
```
jyemin: This is not true.  Take a look at QueryBuilder in 2.x and Filters in 3.x.

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+```
+
+In this case, we're instructing Morphia to add a filter using [$gte]({{< docsref "reference/operator/query/gte/" >}}).  This would result
+ in a query that looks like this:
+ 
+```javascript
+{ price: { $gte: 1000 } }
+```
+
+The list of supported filter operations can be found in the [FilterOperator]({{< srcref
```
jyemin: I think it's worth listing them all in a table.

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+{{% /note %}}
+
+## Complex Queries
+
+Of course, queries are usually more complex than single field comparisons.  Morphia offers both `and()` and `or()` to build up more 
+complex queries.  An `and` query might look something like this:
+
+```java
+q.and(
+    q.criteria("width").equal(10),
```
jyemin: The criteria method threw me.  Why not the field method, which was just mentioned above?

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+    q.criteria("height").equal(1)
+);
+```
+
+An `or` clause looks exactly the same except for using `or()` instead of `and()`, of course.  For these clauses we use the `criteria()`
```
jyemin: Why is it that way (criteria instead of field for and/or)?

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+result in an error.
+{{% /note %}}
+
+## Complex Queries
+
+Of course, queries are usually more complex than single field comparisons.  Morphia offers both `and()` and `or()` to build up more 
+complex queries.  An `and` query might look something like this:
+
+```java
+q.and(
```
jyemin: I thought there was a fluent way of doing and as well.  Something like

    query.field("price").greaterThanOrEq(1000).and("quantity").lessThan(42)

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+    q.criteria("height").equal(1)
+);
+```
+
+An `or` clause looks exactly the same except for using `or()` instead of `and()`, of course.  For these clauses we use the `criteria()` 
+method instead of `field()` but it is used in much the same fashion.  `and()` and `or()` take a [`varargs`](https://docs.oracle
+.com/javase/8/docs/technotes/guides/language/varargs.html) parameter of type `Criteria` so you can include as many filters as necessary.
+
+
+## Text Searching
```
jyemin: It's inconsistent that Text Searching has so much more detail than the other sections.  What's the rationale?

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+"good" in any document.  We created four such documents and our query returns exactly those four.
+
+## Other Query Options
+
+There is more to querying than simply filtering against different document values.  Listed below are some of the options for modifying 
+the query results in different ways.
+
+### Projections
+
+[Projections]({{< docsref "reference/method/db.collection.find/#projections" >}}) allow you to return only a subset of the fields in a
```
jyemin: http://docs.mongodb.org/manual/tutorial/project-fields-from-query-results/ might be a better link

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+
+[Projections]({{< docsref "reference/method/db.collection.find/#projections" >}}) allow you to return only a subset of the fields in a 
+document.  This is useful when you need to only return a smaller view of a larger object.  Borrowing from the [unit tests]({{< srcref 
+"morphia/src/test/java/org/mongodb/morphia/TestQuery.java" >}}), this is an example of this feature in action:
+
+```java
+ContainsRenamedFields user = new ContainsRenamedFields("Frank", "Zappa");
+getDs().save(user);
+
+ContainsRenamedFields found = getDs().find(ContainsRenamedFields.class).retrievedFields(true, "first_name").get();
```
jyemin: The retrievedFields method is the key thing here, but it's buried amidst the rest of the code.  It's also not clear how to retrieve more than one field.  Does it take varargs?  The first argument also deserves an explanation.

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+```
+There's a lot to unwind here so let's take it one item at a time.
+
+1.  This tells Morphia to make sure that any entity [configured](/guides/annotations/#entity) to use a capped collection has its collection 
+created correctly.
+1.  Periodically add documents to the collection
+1.  [Awaitility](https://github.com/jayway/awaitility) is handy utility to wait for a certain condition to be met while also timing out 
+if things take too long.  For more information, please see its [website](https://github.com/jayway/awaitility).
+1.  Since this `Iterator` is backed by a tailable cursor, `next()` will block until a new item is found.  If your application can't block
+ on `next()`, `hasNext()` works as you'd expect an Iterator to.
```
jyemin: end with a new line

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,250 @@
...
+Ordering the results of a query is done via [`Query.order(String)`](/javadoc/org/mongodb/morphia/query/Query.html#order-java.lang.String-)
+.  The javadoc has complete examples but this String consists of a list of comma delimited fields to order by.  To reverse the sort order
+ for a particular field simply prefix that field with a `-`.  For example, to sort by age (youngest to oldest) and then income (highest 
+ to lowest), you would use this:
+ 
+```java
+query.order("age,-income");
+```
+
+### Tailable Cursors
```
jyemin: This example if overly complicated by the executor server and Awaitility.  Can you simplify it?

morphia/src/main/java/org/mongodb/morphia/query/Query.java
```diff
@@ -15,166 +15,142 @@
  * @author Scott Hernandez
```
jyemin: This is a mess.  What does this have to do with the reference documentation?  And why are the diffs so difficult to parse?  Did you move all the methods around?

docs/reference/content/guides/querying.md
evanchooly: doh.  you're right.  I always forget about QueryBuilder.

docs/reference/content/guides/querying.md
evanchooly: I debated that but wasn't sure how best to represent that.  I'll play with some options.

docs/reference/content/guides/querying.md
evanchooly: Sadly, no, though I have some ideas on how to make this bit of the API a little less awkward.  Having to break out in to a different "Criteria" API is a little weird.

docs/reference/content/guides/querying.md
evanchooly: Historical reasons.  I'm not sure really other than field().someOp() returns a Query and I don't think the builder is quite smart enough to make that transition.  Yet, anyway.  I have ideas...

docs/reference/content/guides/querying.md
evanchooly: The example code is a bit more but overall the section doesn't feel any bigger to me.  I used a larger code sample just to show a bit more context when running a text search. There certainly wasn't any explicit decision to do anything terribly different with this section.

docs/reference/content/guides/querying.md
evanchooly: Good call.  Done.

docs/reference/content/guides/querying.md
evanchooly: done

docs/reference/content/guides/querying.md
evanchooly: Sure.  I'll take out all "window dressing" around it since if readers want to see it all they can look at the test its taken from.

morphia/src/main/java/org/mongodb/morphia/query/Query.java
evanchooly: I mentioned in the PR comments that this file was simply reorganized and could be ignored.

jyemin: Sorry, missed that.  Next time leave it out of the PR.

docs/reference/content/guides/querying.md
evanchooly: pushed an update

docs/reference/content/guides/querying.md
jyemin: I'm able to do this:

    ds.createQuery(UserMentionEntity.class).field("x").lessThan(5).field("y").greaterThan(4);

Does it not do what I expect?

evanchooly: That will yes.  That's the implicit `and` than any query document ends up with.  But there isn't a terribly "fluent" API with an explicit `and` or `or`.

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,258 @@
...
+    found.add(tail.next());                                                    // #2
+}
+```
+There are two things to note about this code sample:
+
+1.  This tells Morphia to make sure that any entity [configured](/guides/annotations/#entity) to use a capped collection has its collection 
+created correctly.  If the collection already exists and is not capped, you will have to manually [update]({{< docsref 
+"core/capped-collections/#convert-a-collection-to-capped" >}}) your collection to be a capped collection.
+1.  Since this `Iterator` is backed by a tailable cursor, `next()` will block until a new item is found.  If your application can't block
+ on `next()`, `hasNext()` works as you'd expect an Iterator to.  In this version of the unit test, we tail the cursor and pull out
```
jyemin: What does "`hasNext()` works as you'd expect an Iterator to" mean in this context?  Does it always return true and not block?  Or does it also block?  Remember we dealt with this in the Java driver a while back...

docs/reference/content/guides/querying.md
evanchooly: But I *will* add a note to that effect.  That's a good idea.

docs/reference/content/guides/querying.md
evanchooly: I don't recall that actually.  Have a URL/log handy?  I can tweak that language if you'd like.  It's a bit awkward for sure.

jyemin: https://jira.mongodb.org/browse/JAVA-1255

evanchooly: interesting.  *reading*

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,268 @@
...
+datastore.save(new Greeting("guten Nacht", "german"));
+
+List<Greeting> good = datastore.createQuery(Greeting.class)
+                             .search("good")
+                             .order("_id")
+                             .asList();
+Assert.assertEquals(4, good.size());
+```
+
+As you can see here, we create `Greeting` objects for multiple languages.  In our test query, we're looking for occurrences of the work
```
jyemin: "of the word"

morphia/src/main/java/org/mongodb/morphia/query/Query.java
jyemin: Not to beat a dead horse, but why are you doing this?  It's going to make diff-ing this file very painful in the future.

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,268 @@
...
+    found.add(tail.next());                                                    // #2
+}
+```
+There are two things to note about this code sample:
+
+1.  This tells Morphia to make sure that any entity [configured](/guides/annotations/#entity) to use a capped collection has its collection 
+created correctly.  If the collection already exists and is not capped, you will have to manually [update]({{< docsref 
+"core/capped-collections/#convert-a-collection-to-capped" >}}) your collection to be a capped collection.
+1.  Since this `Iterator` is backed by a tailable cursor, `next()` will block until a new item is found.  If your application can't block
+ on `next()`, checking the return value of `hasNext()` periodically would allow your application to do other things while the cursor
```
jyemin: So are you saying that hasNext() can return false, then later return true?  That's certainly not the case for the Java driver, so I don't see how it could be true for Morphia.

docs/reference/content/guides/querying.md
```diff
@@ -0,0 +1,268 @@
...
+
+datastore.save(new Greeting("guten Morgen", "german"));
+datastore.save(new Greeting("guten Tag", "german"));
+datastore.save(new Greeting("guten Nacht", "german"));
+
+List<Greeting> good = datastore.createQuery(Greeting.class)
+                             .search("good")
+                             .order("_id")
+                             .asList();
+Assert.assertEquals(4, good.size());
```
jyemin: This is the only use of Assert in the docs.  I think it should go.

evanchooly: I use it also in the snippet below.  I can remove them but I'm trying to at least show a comparison against the expected values...

docs/reference/content/guides/querying.md
evanchooly: oh, i see what you're getting at. I thought I saw it behave that way in an early version of the test but I'll double check.  You're probably right, though.

Issue #789, opened 2015-05-29
Add an option for caching the results of Class.forName in DefaultCreator


rlodge:
During performance testing we discovered that calls to Class.forName were taking a large amount of time relative to our expectations.  We'd like to have an option for caching the results of the class name lookup so that the forName method is only called on new classes.

evanchooly:
That happens in a number of places in Morphia.  Any particular hotspot giving you problems?

rlodge:
DefaultCreator.  We've made a change and seen a 10-12% performance improvement, at least in our case.  I'm preparing a pull request.

evanchooly:
excellent.  I look forward to seeing it.  it won't make the 1.0 cut but we can do a 1.0.1 release soon enough.

rlodge:
Works for me, as long as it's on the roadmap.

PR #790, opened 2015-05-29
Issue 789

rlodge:
Changes related to #789  to add optional caching of class names during mapping, which we found increased performance 10-12% in our use case.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/mapping/DefaultCreator.java
+++ b/morphia/src/main/java/org/mongodb/morphia/mapping/DefaultCreator.java
@@ -15,13 +15,26 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 /**
  * @author ScottHernandez
  */
 public class DefaultCreator implements ObjectFactory {
+
     private static final Logger LOG = MorphiaLoggerFactory.get(DefaultCreator.class);
+
+    private Map<String, Class> classNameCache = new ConcurrentHashMap<String, Class>();
+
+    private MapperOptions options = null;
+
+    public DefaultCreator() {
+    }
+
+    public DefaultCreator(MapperOptions options) {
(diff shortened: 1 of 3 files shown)


evanchooly:
Trying to adhere to semantic version, this new feature would have to wait for 1.1 (which hopefully will be the next release anyway)

evanchooly:
So I've pulled this and tweaked it slightly.  The commit is above.  Let me know what you think.  The gist of it trying to reuse the `DefaultCreator` (`ObjectFactory` actually) from the `MappingOptions` we get from the `Mapper`.  At one point during validation, we were creating a new `DefaultCreator` to validate certain bits.  This change pushes the actual `ObjectFactory` down in to the validators so we can reuse this cache.

Also, I changed the `HashMap` to a nonblocking map so as to not incur the locking penalty.  The downside is potentially double writing a `String`/`Class` mapping but that's essentially a no-op the second time around.  The plus side is that all the reads don't need to lock what will eventually be a steady state map.

What do you think?

rlodge:
The changes to push the `ObjectCreator` down into the `MappingValidator` and the `VersionMisuse` validation look good to me.

I'm concerned that the change from `ConcurrentHashMap` to `HashMap` may cause problems.  According to the [API docs for HashMap](https://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html), the HashMap implementation just isn't safe to use in locations where it might be written by multiple threads, so I'm concerned that the change might blow up under load even in a situation where it's passing in unit tests or working ok under light duty.

My understanding is that `ConcurrentHashMap` is non-blocking for most reads, and seems to be [the fastest of the other available options](http://java.dzone.com/articles/java-7-hashmap-vs) in the core API, so I guess unless there's another thread-safe option I'd prefer if we leave the map as a `ConcurrentHashMap`.

evanchooly:
That's fair.  I just reread the docs for CHM and it looks like you're right about the reads.  I'll change that back then and merge this PR.  Thanks for the patch and the feedback.

Issue #784, opened 2015-05-22
NullPointerException mapping class implementing an interface


apara:
Using Morphia 1.0.0-rc0:

Invoking Morphia.map() on a class which implements an interface produces a NullPointerException because:

in ReflectionUtils.getParameterizedClass:

    public static Class getParameterizedClass(final Class c, final int index) {
        final TypeVariable[] typeVars = c.getTypeParameters();
        if (typeVars.length > 0) {
            final TypeVariable typeVariable = typeVars[index];
            final Type[] bounds = typeVariable.getBounds();

            final Type type = bounds[0];
            if (type instanceof Class) {
                return (Class) type; // broke for EnumSet, cause bounds contain
                // type instead of class
            } else {
                return null;
            }
        } else {
            final Type superclass = c.getGenericSuperclass();
            if (superclass instanceof ParameterizedType) {
                final Type[] actualTypeArguments = ((ParameterizedType) superclass).getActualTypeArguments();
                return actualTypeArguments.length > index ? (Class<?>) actualTypeArguments[index] : null;
            } else if (!Object.class.equals(superclass)) {
                return getParameterizedClass((Class) superclass);
            } else {
                return null;
            }
        }
    }

This line: 

    final Type superclass = c.getGenericSuperclass(); 

when executed on an interface results in superclass being null which causes a consequent invocation during execution of 

    getParameterizedClass((Class) superclass) 

to fail with NPE.

markmgordon:
Please target this fix for the next 1.0.x release.  This is an old bug and is preventing us from upgrading.  It looks like trishagee has a fix for you.

evanchooly:
@trishagee how is this looking in your fork?  there's no PR just yet...

PR #792, opened 2015-06-04
Added a test to prove the problem of issue #784

trishagee:
I'm making an assumption that this is the real expected behaviour.


evanchooly:
so no PR on the actual fix?  seems like it's there in your fork...

PR #837, opened 2015-08-30
Logging Log4J

danielfelgar:
A Factory to logging with Log4J.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ bin
 out/
 docs/landing/public
 docs/reference/public
+/build/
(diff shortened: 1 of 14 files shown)


evanchooly:
There are a number changes here unrelated to logging.  Can you clean those up?  From what I can tell so far this looks fine but there's a lot of cruft to dig through.

evanchooly:
Rather than adding support for log4j directly, why not use the slf4j/log4j bridge?  This is what the java driver does, too, and it's been working well for quite some time.

evanchooly:
Given this PR's age and the existence of a workaround, I'm closing this for now.  It can be revisited should something else come up.

PR #845, opened 2015-09-09
correct query validation exception messages fixes #844

### LINKED ISSUE #844: Incorrect validation exception messages in QueryValidator
In QueryValidator's validateQuery method, when a mapped field is not found for a given part, the exception message refers to the original class being mapped rather than the class being used to look up the field:

```
if (mf == null) {                                                                                              
    throw new ValidationException(format("The field '%s' could not be found in '%s' while validating - %s; if "
                                         + "you wish to continue please disable validation.", part,            
                                         clazz.getName(), prop                                                 
                                        ));                                                                    
}                                                                                                              
```

Further along, when checking for queries into `@Reference` or `@Serialized` fields, the exception message is:
```
format("Can not use dot-notation past '%s' could not be found in '%s' while"
       + " validating - %s", part, clazz.getName(), prop))
```
which looks like a copy-paste typo. 

I have a branch off of master on my fork which fixes these issues; should I create a pull request?

shils:


evanchooly:
Could you add a test or two showing the case in error?  There are often multiple code paths to any point in code and we need to make sure we're addressing them correctly.

shils:
Sure, would you prefer an amended commit or a separate one?

evanchooly:
Amended is fine. Thanks so much.

*--------------------------------*

*{ *name     : *"**Justin Lee**", *
  title    : *"*Software Engineer*"*,
  twitter  : *"*@evanchooly <http://twitter.com/evanchooly>*"*,
  web      : *[* "mongodb.com <http://www.mongodb.com>", *"*antwerkz.com*"*
*]*,
  location : *"*New York, NY*"* *}*

shils:
Amended with tests for both cases.

evanchooly:
Looks great.  Thanks for the patch.  good catch.

PR #848, opened 2015-09-16
Allow for multiple Datastores from one Mapper

evanchooly:
@scotthernandez @jyemin @trishagee

This change removes any coupling between Mappers and Datastores. Even with the DatastoreProviders there were numerous places where hard links between specific Mappers and Datastores showed up. These links break users wishing to share a mapping between Datastores. I've left in MapperOptions#setDatastoreProvider(), e.g., to limit any compiler breakage and logged a warning should any one call it.

In general, I feel like this is the correct fix but it does make me a little nervous. There were tests explicitly checking serializability of various DatastoreProvider implementations. I can't see any internal need for such a requirement and have disabled those tests (to be deleted). However, there had to have been a reason for adding those tests in the first place so we need to determine those reasons and if they still apply. (My intuition says no, but ...)

If we need to sit down in person and talk things through that's fine or we can just discuss changes via comments.


evanchooly:
ping

Issue #614, opened 2014-05-14
Add support for sub-packages when using mapPackage


heyarny:
Hi,
it would be great to let mapPackage(package) search within sub-packages in case you have larger count of models in separate packages instead of just one.

PR #850, opened 2015-09-18
Added support for mapping sub-packages.

wittyameta:
Support for mapping sub-packages is added.
Refers to #614

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/Morphia.java
+++ b/morphia/src/main/java/org/mongodb/morphia/Morphia.java
@@ -225,8 +225,19 @@ public class Morphia {
      * @return the Morphia instance
      */
     public synchronized Morphia mapPackage(final String packageName, final boolean ignoreInvalidClasses) {
+        return mapPackage(packageName, false, !mapper.getOptions().isMapSubPackages());
+    }
+    /**
+     * Tries to map all classes in the package specified.
+     *
+     * @param packageName          the name of the package to process
+     * @param ignoreInvalidClasses specifies whether to ignore classes in the package that cannot be mapped
+     * @param ignoreSubPackages    specifies whether to ignore sub-packages during mapping
+     * @return the Morphia instance
+     */
+    private synchronized Morphia mapPackage(final String packageName, final boolean ignoreInvalidClasses, final boolean ignoreSubPackages) {
         try {
-            for (final Class clazz : ReflectionUtils.getClasses(packageName)) {
+            for (final Class clazz : ReflectionUtils.getClasses(packageName, ignoreSubPackages)) {
                 try {
                     final Embedded embeddedAnn = ReflectionUtils.getClassEmbeddedAnnotation(clazz);
                     final Entity entityAnn = ReflectionUtils.getClassEntityAnnotation(clazz);
@@ -255,7 +266,7 @@ public class Morphia {
      * @return this
      */
     public Morphia mapPackageFromClass(final Class clazz) {
-        return mapPackage(clazz.getPackage().getName(), false);
+        return mapPackage(clazz.getPackage().getName(), false, true);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -40,8 +40,8 @@ configure(subprojects.findAll { it.name != 'util' }) {
     apply plugin: 'optional-base'
 
     group = 'org.mongodb.morphia'
-    sourceCompatibility = JavaVersion.VERSION_1_6
-    targetCompatibility = JavaVersion.VERSION_1_6
+    sourceCompatibility = JavaVersion.VERSION_1_8
```
evanchooly: 1.6 is the maximum JVM version we're requiring at this point.  Many of our users are still on older versions and bumping this would break them.

wittyameta: Alright. I wanted to use java.nio. But even that requires Java 1.7

morphia/src/main/java/org/mongodb/morphia/Morphia.java
```diff
@@ -255,7 +266,7 @@ public synchronized Morphia mapPackage(final String packageName, final boolean i
      * @return this
      */
     public Morphia mapPackageFromClass(final Class clazz) {
-        return mapPackage(clazz.getPackage().getName(), false);
+        return mapPackage(clazz.getPackage().getName(), false, true);
```
evanchooly: I agree with the default of ignoring subpackages to avoid an invisible change of functionality but this feels a bit asymmetric.  I think the option to map subpackages should be on MapperOptions rather than adding another opaque method parameter.

morphia/src/main/java/org/mongodb/morphia/utils/ReflectionUtils.java
```diff
@@ -558,9 +566,13 @@ public static Entity getClassEntityAnnotation(final Class c) {
                 jarEntry = jarFile.getNextJarEntry();
                 if (jarEntry != null) {
                     String className = jarEntry.getName();
-                    if (className.endsWith(".class") && getPackageName(className).equals(packageName)) {
-                        className = stripFilenameExtension(className);
-                        classes.add(Class.forName(className.replace('/', '.'), true, loader));
+                    if (className.endsWith(".class")) {
+                        String classPackageName = getPackageName(className);
+                        if ((ignoreSubPackages && classPackageName.equals(packageName)) || (!ignoreSubPackages && isSubPackage(
+                                classPackageName, packageName))) {
```
evanchooly: the first clause need only check equals().  checking ignoreSubPackages twice is a bit redundant.

morphia/src/main/java/org/mongodb/morphia/utils/ReflectionUtils.java
```diff
@@ -594,6 +609,20 @@ public static Entity getClassEntityAnnotation(final Class c) {
         return classes;
     }
 
+    private static List<String> getFileNames(final List<String> fileNames, final Path directory, final String packageName, final boolean
+            ignoreSubPackages) throws IOException {
+        try (DirectoryStream<Path> stream = Files.newDirectoryStream(directory)) {
```
evanchooly: Java 7 and Java 8 features we can't use unfortunately.

morphia/src/main/java/org/mongodb/morphia/Morphia.java
wittyameta: While mapping a class (instead of package), shouldn't the subpackage mapping always be ignored?
Or do you mean that I should remove the mapSubPackage(...) method and include the functionality in the existing mapPackage(...)

evanchooly: You're probably right on this method.  But, yes, in general I think the option to map subpackages should live on MapperOptions.

wittyameta: Added the option in MapperOptions.

wittyameta: Hey @evanchooly, do these changes look OK to you?

wittyameta: @evanchooly any update on this?

wittyameta: All commits squashed into a single commit.

evanchooly:
Slight tweaks but it's in.  Thanks for the PR.

PR #851, opened 2015-09-24
Missing reference in collection should trigger LazyReferenceFetchingException

agherardi:


evanchooly:
I see what you're doing there and it looks correct.  Could you add a test case to confirm it and I'll merge it all in.

agherardi:
I added a unit test as requested. Can this pull request be merged now?

evanchooly:
Slightly tweaked but merged.  Thanks for the patch.

Issue #858, opened 2015-10-19
Projection constructor


musay:
What's the purpose of ```projection(String field, Projection projection, Projection... subsequent) ``` over ```projection(String field, Projection... projections)``` ?

I've been thinking second form of that is more comfortable. Let say we have a list of projections to apply.

There is just one way to do this, but not efficient;
```
List<Projection> projections = ...
List<Projection> remainings = projections;
remainings.remove(0);
...
pipeline.project(
    projection("_id").suppress(),
    projection("blabla", projections.get(0),  remainings.toArray(new Projection[remainings.size()]))
);
```
Which constructor of Projection suits me? I think this could be very nice.
```
pipeline.project(
    projection("_id").suppress(),
    projection("blabla", projections.toArray(new Projection[projections.size()]))
);
```

trishagee:
The idea of having a single parameter followed by varargs is that this way the API can enforce having *at least* one projection passed in.  Using only varargs, it's perfectly valid to call the projection method without a single projection parameter, which is not correct. It's uncomfortable to use with a list of projections, true, but maybe a better solution might be to allow a list, rather than to take away the compile-time safety of the current approach.

PR #859, opened 2015-10-19
Update Projection.java

musay:
#858


evanchooly:
This is a breaking API change and so can't be merged.  A better change would be to overload the method to take a List<Projection> and validate a length of at least one.  That change should be an easy decision to merge but I can't merge this one as is.

PR #868, opened 2015-11-05
Fixed parameter not empty assertion for "reduce"

ddossot:
It seems the issue was due to a copy/paste error.


evanchooly:
Good catch.  Thanks for the PR.

PR #891, opened 2016-02-11
fix testOut in AggregationTest

mich8bsp:
The order should be Dante first and then Homer. Also, checking book list for both authors

### DIFF
--- a/morphia/src/test/java/org/mongodb/morphia/aggregation/AggregationTest.java
+++ b/morphia/src/test/java/org/mongodb/morphia/aggregation/AggregationTest.java
@@ -162,6 +162,10 @@ public class AggregationTest extends TestBase {
                                             .out(Author.class, options);
         Assert.assertEquals(2, getDs().getCollection(Author.class).count());
         Author author = aggregate.next();
+        Assert.assertEquals("Dante", author.name);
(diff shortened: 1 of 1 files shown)


evanchooly:
This test is currently already passing locally, and jenkins and travis.  are you seeing a failure here?

mich8bsp:
I did see a failure, for MongoDB 3.2.1. Since you said the test is passing, I went back and ran it against 3.0.9 and it passed. I guess something in the order of the cursor iterator changed between the versions. I guess I shouldn't use 3.2.1 with morphia relying on java driver 3.0.2

evanchooly:
I'm running tests against 3.2.1 just fine.  That's odd.  What OS?

mich8bsp:
Win10 64-bit... Weird that we're getting different results, I'll try to look into it...

a-h:
I got the same failing test and fixed it in pretty much the same way, except that I added an order to the aggregation to make sure that Dante always comes out before Homer.

evanchooly:
The plan is for 1.4 to make significant updates to aggregation.  I'm going to close this for now with the understanding that the underlying issue will be properly cleaned up with the finished aggregation API.

PR #892, opened 2016-02-11
Update quick-tour.md

a-h:
Fixed broken link, the current link goes to:

https://github.com/mongodb/morphia/blob/master/morphia/src/test/examples/java/org/mongodb/morphia/examples/QuickTour.java

The fixed one goes to:

https://github.com/mongodb/morphia/blob/master/morphia/src/examples/java/org/mongodb/morphia/example/QuickTour.java


evanchooly:
Good catch.  Thanks

PR #895, opened 2016-02-15
Added limit(1) to exist query

Andy-Voigt:


evanchooly:
What exactly is this trying to fix?  This doesn't really change any behavior.  If you look at the [getKey()](https://github.com/mongodb/morphia/blob/dd828e5f9b52bf470c0db023297fed748c4c3c17/morphia/src/main/java/org/mongodb/morphia/query/QueryImpl.java#L224-L224) implementation, you'll see that a limit of 1 is already being transparently applied.

Issue #905, opened 2016-03-01
The $elemMatch operator matches against all fields within the class


a-h:
Use of the hasThisElement filter creates a query which compares all fields in the object regardless.

For example, given an document "photo" with a subdocument of keywords which has two fields ("keyword" and "score"):

```
public static class PhotoWithKeywords {
    @Id
    private ObjectId id;
    @Embedded
    private List<Keyword> keywords = Arrays.asList(new Keyword("california"), new Keyword("nevada"), new Keyword("arizona"));

    public PhotoWithKeywords() {
    }

    public PhotoWithKeywords(final String... words) {
        keywords = new ArrayList<Keyword>(words.length);
        for (final String word : words) {
            keywords.add(new Keyword(word));
        }
    }
}

@Embedded(concreteClass = Keyword.class)
public static class Keyword {
    private String keyword;
    private int score = 12;

    protected Keyword() {
    }

    public Keyword(final String k) {
        this.keyword = k;
    }

    public Keyword(final String k, final int score) {
        this.keyword = k;
        this.score = score;
    }
}
```

This code:

```
Query<PhotoWithKeywords> query = getDs().find(PhotoWithKeywords.class)
                                        .field("keywords")
                                        .hasThisElement(new Keyword("Scott"));
```

Produces a query similar to the following. 

```
db.collection.find({ "keywords": { $elemMatch: { "keyword": "Scott", "score": 12 } } })
```

In many circumstances, I don't care what the score is, I just want to find documents which have a keyword of Scott in them, i.e. this is an overly broad search, given that in many cases, a subdoment $elemMatch will only be searching for an id value, or subset of data.

It would be better to be able to limit the fields which are compared to support this use case.

Issue #906, opened 2016-03-01
The $elemMatch operator cannot be negated


a-h:
The hasThisElement filter cannot be negated. This would be useful for finding a document which is missing a subdocument.

For example, finding all blog posts which don't have a comment from a user called "Sean".

PR #907, opened 2016-03-01
hasThisElement() - Add ability to use $not and limit field comparison for $elemMatch queries

a-h:
Created to resolve issues #905 and #906.

- The $elemMatch operator cannot be negated
- The $elemMatch operator matches against all fields within the class

I was unable to use Morphia to push a value into an array of embedded documents if the embedded document was missing. This was due to encountering the lack of a $not capability in the hasThisElement filter and the fact that the hasThisElement filter creates a query which compares all values in the val parameter against the MongoDB document.

This pull request creates the ability to negate the hasThisElement method and limit the fields which are compared.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/aggregation/GeoNear.java
+++ b/morphia/src/main/java/org/mongodb/morphia/aggregation/GeoNear.java
@@ -154,12 +154,12 @@ public final class GeoNear {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


evanchooly:
I have a few tweaks (e.g. renamed the boolean variant to `doesNotHaveThisElement` to avoid the magical and confusing `boolean` argument) but I'm in the process of merging.

PR #916, opened 2016-03-31
Added type safety to the GeometryShapeConverter.

trishagee:
This reduces the potential of odd subtle bugs and exposed the fact that there was a call to encode something that didn't need encoding.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/geo/GeometryShapeConverter.java
+++ b/morphia/src/main/java/org/mongodb/morphia/geo/GeometryShapeConverter.java
@@ -35,8 +35,8 @@ public class GeometryShapeConverter extends TypeConverter implements SimpleValue
     }
 
     @Override
-    public Object decode(final Class<?> targetClass, final Object fromDBObject, final MappedField optionalExtraInfo) {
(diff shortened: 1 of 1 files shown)


morphia/src/main/java/org/mongodb/morphia/geo/GeometryShapeConverter.java
```diff
@@ -50,20 +50,24 @@ public Object encode(final Object value, final MappedField optionalExtraInfo) {
...
+        } else {
+            List<Geometry> decodedObjects = new ArrayList<Geometry>();
+            for (final Object objectThatNeedsDecoding : mongoDBGeometry) {
+                // MongoDB geometries are lists of lists of lists...
+                decodedObjects.add(decodeObject((List) objectThatNeedsDecoding,
+                                                geometryFactories.subList(1, geometryFactories.size())));
             }
+            return factory.createGeometry(decodedObjects);
         }
-        return getMapper().getConverters().encode(fromDBObject);
```
trishagee: This call was unnecessary, and probably wrong - calling encode from inside the decode method was suspicious, and in fact all you need is to add the last `fromDBObject` to the `decodedObjects` list

PR #930, opened 2016-05-06
Added @IdGetter annotation for method that returns the ID. Use it to optimize lazy loading

agherardi:
Say an entity contains a lazy-load reference to another entity. The application has read the first entity from the database and now it wants to retrieve the ID of the 2nd entity - just the ID. In this case, there is no need for the lazy-load proxy to read the whole 2nd entity from the database, since the ID is already contained in the first entity - hidden inside the lazy load proxy.

This optimization is very useful when building a RESTful webservice whose GET replies contain hyperlinks. When I implement the GET on the first entity, to build the hyperlink to the 2nd entity I only need the 2nd entity's ID. I can avoid the unnecessary database read, I can avoid an N+1 SELECT kind of issue.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/annotations/IdGetter.java
+++ b/morphia/src/main/java/org/mongodb/morphia/annotations/IdGetter.java
@@ -8,6 +8,9 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
+/**
+ * This annotation allows the lazy-load proxy to return the ID of a referenced entity without reading the reference from the database.
+ */
 @Documented
 @Inherited
 @Retention(RetentionPolicy.RUNTIME)
--- a/morphia/src/main/java/org/mongodb/morphia/annotations/IdGetter.java
+++ b/morphia/src/main/java/org/mongodb/morphia/annotations/IdGetter.java
@@ -1,0 +1,17 @@
+package org.mongodb.morphia.annotations;
+
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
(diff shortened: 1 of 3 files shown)


evanchooly:
I would like to see some tests to see this in action.  As it is, i'm not entirely sure how this should be used (and thus documented).  Could you provide some example tests?

agherardi:
{quote}Could you provide some example tests?{quote}

Sure. I did some manual testing. Would you like to see the manual test or are you looking for automated unit tests?

evanchooly:
If you could add some automated testing it'd certainly be the preferred route.  It'd help me understand how this is used and how to document it for others.

agherardi:
Automated unit test added.

agherardi:
A similar feature exists in Hibernate. See
http://stackoverflow.com/questions/2593722/hibernate-one-to-one-getid-without-fetching-entire-object: "A proxy is initialized if you call any method that is not the identifier getter method."

evanchooly:
ok.  I see what you're trying to get at.  I think there's a cleaner way that users can get this for free without an extra annotation.  Essentially intercept the getter for the ID field and return the value (like you're doing) without hydrating the entire lazy proxy.  I'll see if I can't tweak what you've got here...

evanchooly:
OK.  digging in to what you have this is how I was going to have to implement it as well.  We don't have the same mapping model in morphia as we do in another project so the extra annotation is needed.  I had to tweak your unit test slight because Datastore isn't serializable (why didn't that fail for you?) but looks good otherwise.

agherardi:
> why didn't that fail for you?

Not sure. The whole TestLazySingleReference class was at-ignore'd, I removed at-ignore at the class level and added it at the method level to the methods they were failing due to serialization.

Thank you.

evanchooly:
OK.  that make sense then.  Datastore is no longer serializable so that method is fundamentally broken.  Which means I should probably take a long look at those other tests...

Issue #870, opened 2015-11-17
ensureIndexes() throws unhandleable "Index with name: xyz_1" already exists with different options"


andreas-fey:
When opening a datastore, we call the ensureIndexes() on it so be sure all indexes defined by morphia annotations are created.
If annotations get updated in the software over time, e.g. a sparse=true attribute is set, this call will fail on each existing datastore. We would have to delete the old indexes by hand.
It would be great to have an additional parameter in the ensureIndexes() method to define if existing indexes should be overwritten, like already existing in other methods like 

 ensureIndex(Class<T> clazz, String name, String fields, boolean unique, boolean dropDupsOnCreate);

for single collections.

evanchooly:
The tricky part about that feature is you'd either be recreating indexes every time (terribly expensive) or you'd have to deploy one version passing in true for that boolean and then immediately change it back to not pass in true.

thimmwork:
Would it be possible to check whether an index exists and compare the properties to the annotated index options?
- if an index with same parameters exists, do nothing (adds a roundtrip for reading indexes, but cheaper than reindexing every time)
- if an index with different parameters exists, drop it and create with annotated options (expensive only unce, and imho better than an exception)
- if no index exists, create index (same as current behaviour)

PR #942, opened 2016-05-26
#870: ensureIndexes() throws unhandleable "Index with name: xyz_1" already exists with different options"

UlrichCech:
This is a first implemention for this topic... please take a look here, if this could be a solution and if I handled all situations.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
@@ -12,10 +12,13 @@ import com.mongodb.DefaultDBDecoder;
 import com.mongodb.MapReduceCommand;
 import com.mongodb.MapReduceCommand.OutputType;
 import com.mongodb.MongoClient;
+import com.mongodb.MongoCommandException;
 import com.mongodb.MongoException;
 import com.mongodb.ReadPreference;
 import com.mongodb.WriteConcern;
 import com.mongodb.WriteResult;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
 import org.mongodb.morphia.aggregation.AggregationPipeline;
 import org.mongodb.morphia.aggregation.AggregationPipelineImpl;
 import org.mongodb.morphia.annotations.CappedAt;
@@ -259,7 +262,8 @@ public class DatastoreImpl implements AdvancedDatastore {
     public <T> void ensureIndex(final Class<T> clazz, final String name, final String fields, final boolean unique,
                                 final boolean dropDupsOnCreate) {
         ensureIndex(clazz, name, parseFieldsString(fields, clazz, mapper, true, Collections.<MappedClass>emptyList(),
-                                                   Collections.<MappedField>emptyList()), unique, dropDupsOnCreate, false, false, -1);
+                                                   Collections.<MappedField>emptyList()), unique, dropDupsOnCreate,
+                                                   false, false, -1, false);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


UlrichCech:
@evanchooly Could to have a look at the current build? I don't get a clearance of this pull-request, because of errors in the CLIRR-report from a former commit.
There was also a findbugs-error in the Group.class which I wanted to address, perhaps you can have a special look at the Group.class of this pull-request... perhaps, this can be reverted or is obsolete.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -1144,12 +1151,58 @@ protected void ensureIndex(final MappedClass mc, final DBCollection dbColl, fina
...
         LOG.debug(format("Creating index for %s with keys:%s and opts:%s", dbColl.getName(), keys, options));
-        dbColl.createIndex(keys, options);
+        createMongoIndex(dbColl, keys, options);
+    }
+
+    private void createMongoIndex(final DBCollection dbColl, final DBObject keys, final DBObject options) {
+        try {
+            dbColl.createIndex(keys, options);
+        } catch (MongoCommandException ex) {
+            if (options.containsField("dropOutdated") && getBoolean(options, "dropOutdated", false)) {
```
evanchooly: A cast to BasicDBObject is sufficient here and then you'd have access to these methods defined there.

Also, checking contains first is redundant in this case since the default on the get is false.

morphia/src/main/java/org/mongodb/morphia/annotations/Index.java
```diff
@@ -105,6 +105,14 @@
     boolean unique() default false;
 
     /**
+     * Removes an existing index, if the options of the existing index with the same name have changed.
+     *
+     * @deprecated use the {@link IndexOptions} found in {@link #options()}
+     */
+    @Deprecated
+    boolean dropOutdated() default false;
```
evanchooly: This shouldn't be here.  It doesn't make sense to add something that's deprecrated.  New features should go on the new code only.

morphia/src/main/java/org/mongodb/morphia/annotations/Indexed.java
```diff
@@ -76,6 +76,14 @@
     boolean unique() default false;
 
     /**
+     * Removes an existing index, if the options of the existing index with the same name have changed.
+     *
+     * @deprecated use the {@link IndexOptions} found in {@link #options()}
+     */
+    @Deprecated
+    boolean dropOutdated() default false;
```
evanchooly: Again, this value should only be `IndexOptions`

morphia/src/main/java/org/mongodb/morphia/aggregation/Group.java
```diff
@@ -233,7 +234,12 @@ public String getSourceField() {
      * @return the projections for the group
      */
     public Projection[] getProjections() {
-        return projections;
+        if (projections != null) {
```
evanchooly: This doesn't belong in this PR and has been fixed elsewhere.

UlrichCech:
@evanchooly  I don't get the PR green on Travis... it says, that there are 3 test-failures, but I don't know where... if I run the test-suite locally, everything is green.

evanchooly:
master is green too.  try rebasing and force pushing an update to the PR.  Or creating a new one.

UlrichCech:
I try with a new PR

PR #945, opened 2016-05-28
#870: ensureIndexes() throws unhandleable "Index with name: xyz_1" already exists

UlrichCech:
- added a new parameter to IndexOptions for controlling the recreation of the outdated index

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
@@ -12,10 +12,13 @@ import com.mongodb.DefaultDBDecoder;
 import com.mongodb.MapReduceCommand;
 import com.mongodb.MapReduceCommand.OutputType;
 import com.mongodb.MongoClient;
+import com.mongodb.MongoCommandException;
 import com.mongodb.MongoException;
 import com.mongodb.ReadPreference;
 import com.mongodb.WriteConcern;
 import com.mongodb.WriteResult;
+import org.bson.BsonDocument;
+import org.bson.BsonString;
 import org.mongodb.morphia.aggregation.AggregationPipeline;
 import org.mongodb.morphia.aggregation.AggregationPipelineImpl;
 import org.mongodb.morphia.annotations.CappedAt;
@@ -259,7 +262,8 @@ public class DatastoreImpl implements AdvancedDatastore {
     public <T> void ensureIndex(final Class<T> clazz, final String name, final String fields, final boolean unique,
                                 final boolean dropDupsOnCreate) {
         ensureIndex(clazz, name, parseFieldsString(fields, clazz, mapper, true, Collections.<MappedClass>emptyList(),
-                                                   Collections.<MappedField>emptyList()), unique, dropDupsOnCreate, false, false, -1);
+                                                   Collections.<MappedField>emptyList()), unique, dropDupsOnCreate,
+                                                   false, false, -1, false);
     }
 
     @Override
@@ -844,7 +848,7 @@ public class DatastoreImpl implements AdvancedDatastore {
                                 final boolean dropDupsOnCreate) {
         ensureIndex(getCollection(collection), name,
                     parseFieldsString(fields, clazz, mapper, true, Collections.<MappedClass>emptyList(),
-                                      Collections.<MappedField>emptyList()), unique, dropDupsOnCreate, false, false, -1);
+                                      Collections.<MappedField>emptyList()), unique, dropDupsOnCreate, false, false, -1, false);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


UlrichCech:
No chance... the gradle output doesn't show the failing tests in the log from Travis build...

trishagee:
I ran the Gradle build locally and I found the following tests failing:

    
    org.mongodb.morphia.aggregation.AggregationTest > testDateToString FAILED
    org.junit.ComparisonFailure at AggregationTest.java:86
          
    org.mongodb.morphia.indexes.TestIndexOptionAsDropOutdated > testDropOutdatedIndexWhenOptionsHaveChangedWithClassIndex FAILED
    java.lang.AssertionError at TestIndexOptionAsDropOutdated.java:53
                                                                         
    org.mongodb.morphia.indexes.TestIndexOptionAsDropOutdated > testDropOutdatedIndexWhenOptionsHaveChangedWithUnnamedIndex FAILED
    java.lang.AssertionError at TestIndexOptionAsDropOutdated.java:53

UlrichCech:
@trishagee Thank you very much. Can you tell me, which version of MongoDB you are using? I use 3.2.6, but I think, that there is a change in the error message, which I parse in order to get the 'failing' index-name, which must be dropped. In these two testcases, the index is not named via annotation.
If I know the MongoDB version, I try to setup another server with exactly this version in order to reproduce the problem...

The test with "testDateToString" has something to do, at what time you start the test (I ran into failing sometimes, but if I start it later than 2 am in the morning, then it worked)... nevertheless, this is not connected with the PR, because my changes does not touch the Aggregation.

UlrichCech:
Ok, found out, that before MongoDB 2.6, the an index, which exists but with different options was not created, but there was no error. Since 2.6, there is an error.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -1144,12 +1151,37 @@ protected void ensureIndex(final MappedClass mc, final DBCollection dbColl, fina
...
+        } catch (MongoCommandException ex) {
+            final BasicDBObject basicOptions = (BasicDBObject) options;
+            if (basicOptions.getBoolean("dropOutdated", false)) {
+                final BsonDocument errorDocument = ex.getResponse();
+                if (errorDocument.getInt32("code").intValue() == 85) {
+                    String indexName;
+                    if (options.containsKey("name")) {
+                        indexName = basicOptions.getString("name");
+                    } else {
+                        final BsonString errorMessage = errorDocument.getString("errmsg");
```
trishagee: While I assume the tests show that this approach works, adding any logic based on the string of an error message is extremely brittle, and not necessarily correct (for example error messages vary between versions and could also potentially be translated). In an ideal world, we would rely on the error codes, but unfortunately we're not able to do that - error codes are not well documented, and are not consistent between MongoDB versions.

trishagee:
The amount of work you've put into this pull request is clear - you have test coverage, have followed the conventions and worked hard to get the build passing.  

However, in MongoDB libraries like Morphia we have generally tried to avoid implementing strategies for automatically dealing with failure (retries on error, for example) because of the added performance cost that is not always visible to those using the library.  It's generally better practice to highlight the failure to the application using the library (e.g. through specific Exceptions) so developers can deal with the problem in a way that is appropriate for their case.

Ultimately the decision to merge pull requests is @evanchooly's, but I wanted to take the time to explain why code that you've clearly spent time and effort on may not be merged immediately and requires careful consideration from the maintainers.

UlrichCech:
Hi Trisha,
thanks for your explanation.
The problem here is, that if we did not provide such "option" to automatically delete and re-create the index when there is an index with that name but with other options, the user has no real option to deal with this situation correctly. The only information the user has at the moment is a simple text-message, where an index-name is provides. But the user does not know, on which collection this index is, because this information is not given. So the only way to handle this is to iterate over all collections, get the index names and throw away this index, which can result in a much unneccessary dropped indexes. And this is not really possible in production systems, where there are huge collections, which need time to recreate.

I think, to have this "dropOutdated"-option is not other then the "dropDubs"-option. And the "automatic failure-code" only come into play, when there is an error situation, so there is no real performance hit.

Nevertheless, relying on messages or something is always not ideal. Better would be, when Mongo-Java-Driver throws some "IndexAlreadyExistsWithOtherOption"-exception and provide a reliable data structure with index-name and collection-name, so Morphia or the user have exact informations to deal with.

evanchooly:
In point of fact, `dropDups` is a server-side option that's been removed for many of the same reasons this option makes me terribly nervous.  Trisha has it right that correctly identifying this case is tricky.  If the problem is that the message doesn't contain the collection, that can be easily fixed.  I'm still unsure about this PR.  It needs a bit of discussion still, I think.

UlrichCech:
@evanchooly 
< If the problem is that the message doesn't contain the collection>
Ok, but if I get it right, we need to update MongoServer, because the error message is generated by Mongo directly.
But then, it would be better, if the response contains a dedicated error structure, otherwise the user has to deal with string-parsing to identify collection- and index-name.

evanchooly:
Indexing has been significantly refactored so this PR can not be merged as is.  In addition, consensus among the team is that is isn't something that should be in Morphia itself.  As such, I'm going to close out this PR.  Issue #1036 would be an interesting one to tackle if you're looking to contribute.  It'd be very similar to the document validation work.

UlrichCech:
Does the MongoDB-Java-Driver now provides the collection name in addition to the index name in the exception? Then I would make the "workaround" in the business-application.

PR #963, opened 2016-07-13
first pass at an explicit reference type rather than an annotation and a proxy

evanchooly:
This is one way this could look.  I've debated repurposing `Key<T>` for this as they're pretty similar but I'm mildly nervous repurposing something like that.  It _probably_ wouldn't break any applications but it might muddy the waters internally in the Mapper classes.


morphia/src/main/java/org/mongodb/morphia/MorphiaReference.java
```diff
@@ -0,0 +1,123 @@
...
+        ref.collection = collection.getName();
+
+        return ref;
+    }
+
+    @SuppressWarnings("unchecked")
+    T fetch(final Datastore datastore) {
+        if (entity == null) {
+            if (collection != null) {
+                entity = (T) ((DatastoreImpl) datastore).getCollection(collection).findOne(id);
```
jyemin: This looks like a call to DBCollection.findOne, which is going to return a DBObject, not a T

evanchooly: Yep.  Missing test case.  Updated.

jyemin:
This looks like a non-starter as is, since it's not backwards compatible with existing DBRefs already stored in MongoDB.  As in, here's what the test creates:

  > db.Container.findOne()
  {
	"_id" : ObjectId("5787f1336e4d379c2cdbb358"),
	"className" : "org.mongodb.morphia.mapping.references.TestMorphiaReference$Container",
	"reference" : {
		"id" : {
			"className" : "org.mongodb.morphia.testmodel.CompoundId",
			"id" : ObjectId("5787f1336e4d379c2cdbb34d"),
			"name" : "I'm an ID!"
		},
		"type" : "org.mongodb.morphia.mapping.references.TestMorphiaReference$Contained",
		"collection" : "somewherelse"
	},
       ...

Is this what you were going for?

evanchooly:
There would be converters to handle the DBRef -> MorphiaReference mapping.

jyemin:
So this isn't ready to merge, more of an API sketch? 

As an API, I think it would be more flexible if MorphiaReference was a lot dumber -- like DBRef --  
and have no behavior, no genericity, no lazy cache of the entity, and no dependency on Datastore.

That way the reference can be used in multiple scenarios.  With Datastore, with MongoCollection directly, even with the async MongoCollection.

The only weird part is that DBRef contains the collection as a name, so not clear to me what the Datastore API would look like.  If you do something like:

   T Datastore.resolve(Class<T>clazz, MorphiaReference reference)

what does it mean if clazz wouldn't resolve to the same collection as the one in the reference?  Also, DBRef can contain a $db field (which the driver now supports).  What if that doesn't even match the database of the Datastore?

evanchooly:
No, this is certainly a rough sketch.  I'm trying to converge on just have a DBRef + some extra data.  

I've been avoiding the `db` field question for now but it shouldn't be a problem.

evanchooly:
Now serializes to a raw `DBRef` and supports loading old documents in to the new class structure.  We can almost just do away with MorphiaReference and use DBRef directly in the model but that still feels a little dirty to me.

evanchooly:
Any other thoughts?  Is this the direction we *want* to go?  My gut says yes.

evanchooly:
this is getting stale.  i'd like to merge.

evanchooly:
currently reworking this a bit...

Issue #799, opened 2015-06-11
Investigate ordering text search by $score and including the return value


evanchooly:
See [this](https://stackoverflow.com/questions/30709505/include-score-in-morphia-full-text-search/30738588) stackoverflow post for a more complete discussion.  Essentially, it would be nice to be able to include the $score values on the returned entity objects after a text search.

More important, however, is being able to order by those scores.

ctmay4:
Thanks for looking into this.

I think the real issue is to add support for $meta (http://docs.mongodb.org/manual/reference/operator/projection/meta/). 

The name of the property ("score" in my example) is defined by the caller as a map to the $meta field.  In Morphia, that field would need to exist in entity obviously.

```java
    BasicDBObject meta = new BasicDBObject("$meta", "textScore");
    BasicDBObject score = new BasicDBObject("score", meta);
```

ctmay4:
Any chance this issue could be looked at for a future release.  I keep running into it an I have to solve it the same way each time.

Instead of just being able to say

```java
    Query<NdcProduct> q = ...
    q.order(...);
    return q.asList();
```

This code is required to return full text results ordered by score.

1. Build my original query in Morphia

    ```java
    Query<NdcProduct> q = ...
    ```

2.  Extract query information into DBObjects so I can sort by score.

    ```java
    // need to build the score for sorting purposes
    BasicDBObject meta = new BasicDBObject("$meta", "textScore");
    BasicDBObject score = new BasicDBObject("score", meta);

    List<DBObject> results = q.getCollection().find(q.getQueryObject(), score).sort(score);
    ```

3.  Loop over the results and convert them back to my original entities from DBObjects

    ```java
    List<NdcProduct> products = new ArrayList<>();
    for (DBObject obj : results) {
        NdcProduct product = morphia.fromDBObject(ds, NdcProduct.class, obj);

        products.add(product);
    }
    ```

IVANOPT:
@ctmay4 your codes really helps me much. Something I suppose that:

1、A small error in codes (The question you asked in stackoverflow is right: https://stackoverflow.com/questions/30709505/include-score-in-morphia-full-text-search/30738588#30738588) :

```
List<DBObject> results = q.getCollection().find(q.getQueryObject(), score).sort(score);

=>

List<DBObject> results = q.getCollection().find(q.getQueryObject(), score).sort(score).toArray();
```

2、 I suggest to add $meta keyword to full text search function of morphia. @evanchooly

PR #964, opened 2016-07-17
$meta projection operator support in text search query

elogunov:
I'd like to propose the implementation of [$meta](https://docs.mongodb.com/manual/reference/operator/aggregation/meta/#exp._S_meta) projection for Query. I belive, the request is linked with issue #799
The request doesn't contains $meta projection operator support for AgregationPipline. I think it can be done later.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/query/Meta.java
+++ b/morphia/src/main/java/org/mongodb/morphia/query/Meta.java
@@ -1,0 +1,78 @@
+package org.mongodb.morphia.query;
+
+import com.mongodb.BasicDBObject;
+import com.mongodb.DBObject;
+
+/**
+ * Defines $meta expression object
+ * @mongodb.driver.manual reference/operator/aggregation/meta/ $meta
+ */
+public class Meta {
+
+    private static final String META = "$meta";
+
+    /**
+     * Representing specified metadata keyword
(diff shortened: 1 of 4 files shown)


evanchooly:
First, this is great.  Thanks for putting this together.  I had a couple of questions/thoughts I want to run by you and see what you think.

In order to make the projection and sorting work, I saw you had to add a `score` property to `Book` which is unfortunate.  I played around with your PR and came up with a way around that.  In `QueryImpl` on lines 465 and 519, if we pass `false` in the `validateNames` parameter, we can bypass validation for the `$meta` fields and thus not pollute our model these score values.

If you feel strongly about keeping that field in the model for some reason, it should at least be a transient field so that it doesn't get saved back to mongodb if that instance is saved again.

I think bypassing the validation is the correct path but wanted to get your thoughts on it before I committed that tweak.

elogunov:
Hi evanchooly. Thanks for the review. I agree about unnecessary `score` property. I like your fix of this problem. I think it is ok to disable names validation (`validateNames`) for `$meta` object. The implementation looks more flexible after the fix.

evanchooly:
Great.  I wanted to make sure it wasn't just me.  I'll merge and push what I have then.  Thanks again for the PR.

PR #972, opened 2016-07-30
$meta projection operator  support in aggregation pipeline

elogunov:
I would like to propose another enhancement  to support `$meta` expression in aggregation pipeline. The idea of the enhancement is to extract interfaces for `Projection` (ProjectionElement), `Group`(GroupElement) and `Sort`(SortElement) classes. New `MetaAggregation` class implements all that interfaces. So there is easy way to add MetaAggregation (using inner `MetaAggregationBuilder`) into `AggregationPipeline`.
I've noticed that there is another branch for aggregation enchancement, and there can be some confilcts with my commit. But still I decided to perform this PR

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,8 @@
 ext.configDir = new File(rootDir, 'config')
 ext.slf4jVersion = '1.7.5'
+ext.git = find('git')
+ext.majorVersion = version.split('\\.').dropRight(1).join('.') + ( gitVersion() == 'master' ? '-SNAPSHOT' : '')
+ext.latest = gitVersion() == 'master' ? version : latest_release
 
 if (JavaVersion.current().isJava8Compatible()) {
     allprojects {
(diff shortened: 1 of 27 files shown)


evanchooly:
First off, thanks for doing the work here.  Sadly, it looks like you've renamed some types and broke backwards compatibility with existing code.  If you'd like to rebase and try again, I'd be happy to look it over.

elogunov:
Hello. I developed new interfaces (for example `GroupElement`), but existing classes (`Group`) implement these interfaces. And methods in `AgregationPipeline` have to be backward compatible. Can you please correct me if i am mistaken?

evanchooly:
I'm afraid this PR has aged out.  And it appears that some related changes have made their way in already and so perhaps this PR has been (mostly) fixed elsewhere.  I encourage you to take a look at 1.4.0 or better yet `master` and see if your needs are still unmet.

PR #973, opened 2016-08-06
Fix typo (de)

nerro:
Fix small typo:
- `guten Nacht` -> `gute Nacht`

http://www.duden.de/rechtschreibung/Nacht


nerro:
Travis builds were ok ,but I saw in build logs strange error. Something important?
```bash
$ Error executing 'clirr' task. java.lang.ClassNotFoundException: org.apache.tools.ant.Task not found.
```

evanchooly:
That clirr message is longstanding.  I haven't figured out why yet but you don't need to worry about it.

PR #979, opened 2016-08-18
Include an AdvancedDatastore-specific method for aggregation

daveehrenberger:
While `AdvancedDatastore` provides many methods to allow an alternative collection name, the creation of an `AggregationPipeline` is not one of them. This pull request is proposing the simple addition of such a method: `AdvancedDatastore.createAggregation(String, Class<?>);`


evanchooly:
Looks good.  Thanks for the PR.

PR #981, opened 2016-08-22
add document validation

evanchooly:


morphia/src/main/java/org/mongodb/morphia/Datastore.java
```diff
@@ -355,6 +365,12 @@
     DB getDB();
 
     /**
+     * @return the MongoDatabase this Datastore uses
+     * @since 1.3
+     */
+    MongoDatabase getDatabase();
```
rozza: Seems tangential to adding validation

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -250,6 +260,40 @@ public void ensureCaps() {
...
 
     @Override
+    public void enableDocumentValidation() {
+        for (final MappedClass mc : mapper.getMappedClasses()) {
+            Validation validation = (Validation) mc.getAnnotation(Validation.class);
+            if (validation != null) {
+                String collectionName = mc.getCollectionName();
+                CommandResult result = getDB()
+                    .command(new BasicDBObject("collMod", collectionName)
+                                 .append("validator", Document.parse(validation.value()))
```
rozza: Nit.

Seems weird to mix DBObject and Document - why not use `BasicDBObject.parse` ?

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -250,6 +260,40 @@ public void ensureCaps() {
...
+                                 .append("validationLevel", validation.level().getValue())
+                                 .append("validationAction", validation.action().getValue())
+                            );
+
+                if (!result.ok()) {
```
rozza: Question: If there was an existing validation - would it be overwritten? Or would it throw an error?

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -250,6 +260,40 @@ public void ensureCaps() {
...
+                    if (result.getInt("code") == 26) {
+                        try {
+                            ValidationOptions options = new ValidationOptions()
+                                .validator(Document.parse(validation.value()))
```
rozza: As above for Document.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: Mostly because I didn't realize it was there.  I'd used JSON.parse() in the past and it failed on certain docs.  I knew of Document.parse() and went for that.  remove all the Document references.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: It would replace the existing validation.  There's a test showing that.

evanchooly:
@jyemin bump

morphia/src/main/java/org/mongodb/morphia/Datastore.java
```diff
@@ -115,6 +117,14 @@
     void ensureCaps();
 
     /**
+     * Process any {@link Validation} annotations for document validation
```
jyemin: What's the default?  It should be documented.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -78,6 +85,7 @@
     private final Morphia morphia;
     private final MongoClient mongoClient;
     private final DB db;
+    private final String dbName;
```
jyemin: No need for another field. Just get it from the 'db' field.

evanchooly: Yep.  Gone.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.mongodb.WriteConcernException;
+import org.junit.Test;
+import org.mongodb.morphia.entities.DocumentValidation;
+
+import java.util.Date;
+
+import static org.junit.Assert.assertTrue;
+
+public class TestDocumentValidation extends TestBase {
```
jyemin: These tests are not complete and don't assert enough for what they do test.

* No tests for Validation.level
* No tests for Validation.action
* No assertion that the requested validation has been added to a new collection
* No assertion that the requested validation has updated an existing validator on an existing collection

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
```diff
@@ -0,0 +1,46 @@
...
+    @Test
+    public void createValidation() {
+        getMorphia().map(DocumentValidation.class);
+        getDs().enableDocumentValidation();
+
+        try {
+            getDs().save(new DocumentValidation("John", 1, new Date()));
+        } catch (WriteConcernException e) {
+            assertTrue(e.getMessage().contains("Document failed validation"));
```
jyemin: Why is this assertion here?  There's no Morphia-specific code that's being tested.  But see above for assertions that _should_ be here.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -250,6 +259,38 @@ public void ensureCaps() {
     }
 
     @Override
+    public void enableDocumentValidation() {
+        for (final MappedClass mc : mapper.getMappedClasses()) {
```
jyemin: Just to confirm: is the list of mapped classes essentially immutable at this point?

evanchooly: Not in morphia, no.  Classes can be dynamically added/mapped later.  So here, as with indexes and capped collections, classes will have to be explicitly mapped before calling this method.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
evanchooly: This is trying to save a document that doesn't pass validation.  The  'number' field must be > 10 and i'm passing in a 1.

jyemin: Acknowledged.  I was confused by the lack of a fail(...) in the try block.  And by the fact that the driver throws a WriteConcernException for something that has nothing to do with write concern.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
jyemin: I still don't see tests for validation level and action.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
```diff
@@ -0,0 +1,106 @@
...
+        try {
+            database.getCollection("validation").insertOne(rhubarb);
+            fail("Document should have failed validation");
+        } catch (MongoWriteException e) {
+            assertTrue(e.getMessage().contains("Document failed validation"));
+        }
+
+        getMorphia().map(DocumentValidation.class);
+        getDs().enableDocumentValidation();
+        assertEquals(Document.parse(DocumentValidation.class.getAnnotation(Validation.class).value()), getValidator(database));
```
jyemin: Should add this assertion to the createValidation test as well

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        try {
+            getDs().save(new DocumentValidation("John", 1, new Date()));
+            fail("Document should have failed validation");
+        } catch (WriteConcernException e) {
+            assertTrue(e.getMessage().contains("Document failed validation"));
+        }
+
+        getDs().enableDocumentValidation();
```
jyemin: It's not clear to me what this last line is testing.

evanchooly: it's vestigial at this point.  I've removed it.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
evanchooly: The error action is already being tested.  I'm not entirely sure how to test the warn level since i think that just results in a log message.  I think the best I can do is simply validate the 'validator' document in the db.  perhaps that's sufficient.

jyemin: You could test that the validation action is applied properly by creating an annotation where the action is WARN, and then assert that an invalid document is still inserted.

You could test validation level is applied properly by creating an annotation where the level is MODERATE and then assert that an already invalid document is not validated on update.

But it's sufficient to assert that the values for validationLevel and validationAction  in the options field from collection info matches the ones (for both the create collection and the collMod cases)

evanchooly: Yeah, I'm writing those tests now.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
evanchooly: done.

evanchooly:
bump

evanchooly:
if there are no other objects, i'll merge to master.

morphia/src/test/java/org/mongodb/morphia/TestDocumentValidation.java
jyemin: Looks like this was done, right?

jyemin:
What are all 19 commits from August 31?

evanchooly:
rebased against master since it'd been a while.  trying to get this prepped for merge.

evanchooly:
something clearly went wonky so i'm going to close this one and open a new so that the diffs are obvious.

PR #985, opened 2016-08-31
new implementation of $elemMatch support

### LINKED ISSUE #970: $elemMatch multiple criteria support
Correct me if I am wrong but as I can see current Morphia implementation of `$elemMatch` (i.e. `FieldEnd#hasThisElement`) doesn't support [multiple criteria for array elements](https://docs.mongodb.com/manual/tutorial/query-documents/#specify-multiple-criteria-for-array-elements) or [multiple criteria for array of documents](https://docs.mongodb.com/manual/tutorial/query-documents/#specify-multiple-criteria-for-array-of-documents).
So there is no way to create such query using morphia:
`db.restaurants.find({grades : { $elemMatch:{"score":{$gt : 80 , $lt :100}}}})`
[stackoverflow question](https://stackoverflow.com/questions/37650166/how-to-write-the-following-query-in-morphia)

I didn't find any linked issues here.

evanchooly:
This is a replacement implementation of $elemMatch.  The old "hasThisElement" and "doesNotHaveThisElement" are being deprecated as they are rather limited in what they support.  This resolves #970.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
@@ -142,7 +142,7 @@ class FieldEndImpl<T extends CriteriaContainerImpl> implements FieldEnd<T> {
     @Override
     public T hasThisElement(final Object val) {
         Assert.parametersNotNull("val", val);
-        return addCriteria(FilterOperator.ELEMENT_MATCH, val, not);
+        return addCriteria(FilterOperator.ELEMENT_MATCH, val, false);
     }
 
     @Override
--- a/morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
@@ -142,13 +142,13 @@ class FieldEndImpl<T extends CriteriaContainerImpl> implements FieldEnd<T> {
     @Override
     public T hasThisElement(final Object val) {
         Assert.parametersNotNull("val", val);
-        return addCriteria(FilterOperator.ELEMENT_MATCH, val, false);
+        return addCriteria(FilterOperator.ELEMENT_MATCH, val, not);
     }
 
     @Override
     public T elemMatch(final Query query) {
         Assert.parametersNotNull("query", query);
-        return addCriteria(FilterOperator.ELEMENT_MATCH, query, false);
+        return addCriteria(FilterOperator.ELEMENT_MATCH, query, not);
     }
 
     @Override
--- a/morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
@@ -146,6 +146,12 @@ class FieldEndImpl<T extends CriteriaContainerImpl> implements FieldEnd<T> {
     }
 
     @Override
+    public T elemMatch(final Query query) {
+        Assert.parametersNotNull("query", query);
+        return addCriteria(FilterOperator.ELEMENT_MATCH, query, false);
+    }
+
+    @Override
     public T hasThisOne(final Object val) {
         return addCriteria(FilterOperator.EQUAL, val);
     }
--- a/morphia/src/test/java/org/mongodb/morphia/TestQuery.java
+++ b/morphia/src/test/java/org/mongodb/morphia/TestQuery.java
@@ -381,7 +381,7 @@ public class TestQuery extends TestBase {
     }
 
     @Test
-    public void testElemMatchQueryCanBeNegated() {
+    public void testElemMatchVariants() {
         final PhotoWithKeywords pwk1 = new PhotoWithKeywords();
         final PhotoWithKeywords pwk2 = new PhotoWithKeywords("Kevin");
         final PhotoWithKeywords pwk3 = new PhotoWithKeywords("Scott", "Joe", "Sarah");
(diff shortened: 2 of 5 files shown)


morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
```diff
@@ -142,7 +142,13 @@ public T doesNotHaveThisElement(final Object val) {
     @Override
     public T hasThisElement(final Object val) {
         Assert.parametersNotNull("val", val);
-        return addCriteria(FilterOperator.ELEMENT_MATCH, val, false);
+        return addCriteria(FilterOperator.ELEMENT_MATCH, val, not);
```
jyemin: Why did the 3rd param just from false to 'not'?  Is this fixing a bug?

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
```diff
@@ -142,7 +142,13 @@ public T doesNotHaveThisElement(final Object val) {
     @Override
     public T hasThisElement(final Object val) {
```
jyemin: Could 'val' be a DBObject with the query criteria?  Issue 970 implies that it could not.

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
evanchooly: It is.  It's currently not possible to negate this operation because of that hard coded 'false'

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
evanchooly: It can actually.

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
jyemin: Is that bug reported?  If not, it should be so we can track that it's getting fixed.  And if it's not already, the bug fix should be done in a separate commit from the new feature work for elemMatch

morphia/src/test/java/org/mongodb/morphia/TestQuery.java
```diff
@@ -369,37 +391,67 @@ public void testElemMatchQueryCanBeNegated() {
...
+                                                                  .createQuery(Keyword.class)
+                                                                  .field("keyword").equal("Scott"))
+                                                   .asKeyList());
+    }
 
-        keys = getDs().find(PhotoWithKeywords.class)
-                      .field("keywords")
-                      .hasThisElement(new Keyword(14))
-                      .asKeyList();
+    private void validate3(final PhotoWithKeywords pwk1, final PhotoWithKeywords pwk2, final PhotoWithKeywords pwk3, final Mapper mapper,
```
jyemin: validate1/2/3 are not descriptive enough names.   Try to describe the purpose of each one.

morphia/src/test/java/org/mongodb/morphia/TestQuery.java
```diff
@@ -369,37 +391,67 @@ public void testElemMatchQueryCanBeNegated() {
 
         Mapper mapper = getMorphia().getMapper();
 
-        List<Key<PhotoWithKeywords>> keys = getDs().find(PhotoWithKeywords.class)
+        validate(pwk1, pwk2, pwk3, pwk4, mapper, getDs().find(PhotoWithKeywords.class)
```
jyemin: validate and validate2 are not testing what the test method says it is: that elemMatch queries can be negated. Only validate3 is doing that.  So put the validate/validate2 assertions in a separate test.

evanchooly: those are literally just refactorings of existing tests in to methods so i can use both elemMatch queries with the same asserts.  I didn't change what they were testing

jyemin: Yes, I see that the validate methods are a refactoring.  But a bunch more assertions have been added to this test method and I'm suggesting that it would be better to restructure into multiple test methods whose named better describe what it is that is being tested.  And once you've done that, the test names you've chosen will suggest what the validate* methods should be renamed to to reflect their purpose.

Let me know if this doesn't make sense and we can talk about it more.

evanchooly: I was trying to avoid cutting too deeply but I bit the bullet and cleaned that up a bit.  less code now and easier to reason about what's being asserted.

jyemin: I like the new validate method... but there are still assertions that have nothing to do with negation in a test method called testElemMatchQueryCanBeNegated.  That's just confusing, don't you think?

evanchooly: a bit, yeah.  there's only the one negation at the end, really.  i'll brainstorm some new names.  to be honest, i think this test duplicates a number of others and can probably be combined...

evanchooly:
with the method renamed, all else is good @jyemin ?

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
jyemin: Ping on this last comment.  Really should avoid drive-by bug fixes like this.  Open a separate issue and fix it separately.

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
jyemin: Seems like it deserves a comment on issue 970 then, no?

morphia/src/test/java/org/mongodb/morphia/TestQuery.java
jyemin: Looks good now.

jyemin:
I'd still like to see the bug on the existing method reported and the fix pulled out into a separate commit.

morphia/src/main/java/org/mongodb/morphia/query/FieldEndImpl.java
evanchooly: I rolled back that fix but hadn't filed the issue.  I'll do that now.

jyemin:
LGTM

yizhangy:
I find the the error warning message when used the "elemMatch"

```
Query<SelectValue> selectValueQuery = super.getMongoDatastore().createQuery(SelectValue.class);
		selectValueQuery.filter("id =", voteValueId);
		selectValueQuery.filter("checked =", true);
		
	
		
		Query<FeedMemberEntity> query = super.getMongoDatastore().find(FeedMemberEntity.class);
		query.field("feedId").equal(feedId);
		query.field("voteValues").elemMatch(selectValueQuery);
		
		return query.count();
```

My log had
```
[Validation failed: 'Type java.util.List may not be queryable with value '{ query: { "id" : "0" , "checked" : true}  }' with class xyz.morphia.query.QueryImpl', Validation failed: 'Type com.canaan.config.SelectValue may not be queryable with value '{ query: { "id" : "0" , "checked" : true}  }' with class xyz.morphia.query.QueryImpl']
Feb 10, 2019 11:04:44 PM xyz.morphia.query.QueryValidator validateQuery
```
Is this still a bug

evanchooly:
Probably related #1288

yizhangy:
> #1288



> Probably related #1288

thanks refer me to this issue. It is not big issue, just a warning. we could disable the validation when make this query for now. thanks.

PR #989, opened 2016-09-12
Fix error "execCommand == null!"

kimbeast:
When building with gradle on windows or importing using intellij,
"execeCommand ==null" error occurs.
Fix this error

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -153,7 +153,7 @@ def major(fullVersion) {
 def gitVersion() {
     def stdOut = new ByteArrayOutputStream()
     exec {
-        commandLine git, "rev-parse", "--abbrev-ref", "HEAD"
+        commandLine "git", "rev-parse", "--abbrev-ref", "HEAD"
         standardOutput = stdOut
         ignoreExitValue = true
     }


evanchooly:
if `git` is null it's because it doesn't live in one of the two places the build is looking for it.  try adding this to the end of the last line in `find()`  (line 171 in the current [master](https://github.com/mongodb/morphia/blob/e6fc8fb13e8a3985d0c9e66a6bae3060e6fa765f/build.gradle#L171-L171)):

    ?: executable


If that fixes your problem I can push the change to master

kimbeast:
`git` is not null. `git` command is in my system.
When i try to test with only `gitVersion` function, It occurs following error.

` No such property: git for class: org.gradle.process.internal.DefaultExecAction_Decorated`

code is 
```
gitVersion()

def gitVersion() {
    def stdOut = new ByteArrayOutputStream()
    exec {
        commandLine git, "rev-parse", "--abbrev-ref", "HEAD"
        standardOutput = stdOut
        ignoreExitValue = true
    }

    return stdOut.toString().trim()
}
```

**I just fix `git` make to `string`.**

evanchooly:
pushed a different fix

PR #1028, opened 2016-09-27
Index refactoring and collation support

evanchooly:
Adding collation support to indexing meant updating several different code paths because the indexing code in morphia was a bit fractured.  On the way to getting this support added the work evolved in to a refactoring of the indexing code.  All the indexing logic was moved in to `IndexHelper` and the code paths merged.  This part is only the first half of collation support.  With the indexing in place, collation support can then be added to the query/update operations.


jyemin:
I'll wait for the Travis build to be successful.

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
```diff
@@ -0,0 +1,492 @@
...
+    private final MongoDatabase database;
+
+    IndexHelper(final AdvancedDatastore datastore, final Mapper mapper, final MongoDatabase database) {
+        this.datastore = datastore;
+        this.mapper = mapper;
+        this.database = database;
+    }
+
+    @SuppressWarnings("deprecation")
+    protected com.mongodb.client.model.IndexOptions convert(final IndexOptions options, final boolean background) {
```
jyemin: protected seems inappropriate for a package-protected class that should be final

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
```diff
@@ -0,0 +1,492 @@
...
+        if (options.dropDups()) {
```
jyemin: This will cause runtime errors for existing users, right?  What's the justification for that without a deprecation period?  I know the server dropped support for this, but if they didn't upgrade their server...

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
```diff
@@ -0,0 +1,492 @@
...
+                                                 .collationMaxVariable(collation.maxVariable())
+                                                 .collationStrength(collation.strength())
+                                                 .normalization(collation.normalization())
+                                                 .numericOrdering(collation.numericOrdering())
+                                                 .build();
+    }
+
+    @SuppressWarnings("deprecation")
+    private Index synthesizeIndexFromOldFormat(final Index index) {
+        return synthesizeIndex(annotationForMap(IndexOptions.class, toMap(index)), parseFieldsString(index.value()));
```
jyemin: Why isn't this just the identity transform?

morphia/src/main/java/org/mongodb/morphia/mapping/Mapper.java
```diff
@@ -195,6 +195,24 @@ public EntityCache createEntityCache() {
     }
 
     /**
+     * Finds any subtypes for the given MappedClass.
+     *
+     * @param mc the parent type
+     *
+     * @return the list of subtypes
+     */
+    public List<MappedClass> getSubTypes(final MappedClass mc) {
```
jyemin: Unit test this.

morphia/src/main/java/org/mongodb/morphia/mapping/Mapper.java
```diff
@@ -737,7 +755,7 @@ private void addConverters(final MappedClass mc) {
     private MappedClass addMappedClass(final MappedClass mc, final boolean validate) {
         addConverters(mc);
 
-        if (validate) {
+        if (validate && !mc.isInterface()) {
```
jyemin: What's this change for? If it's really necessary it should be unit tested.

morphia/src/main/java/org/mongodb/morphia/utils/IndexType.java
```diff
@@ -11,10 +11,25 @@
...
+
+    /**
+     * Returns the enum instance for the given value
+     *
+     * @param value the value to find
+     * @return the enum instance
+     */
+    public static IndexType fromValue(final Object value) {
+        for (IndexType indexType : values()) {
+            if (indexType.type.equals(value) || indexType.name().equals(value)) {
```
jyemin: Don't allow comparison against both the name and the value.  That's confusing.

morphia/src/main/java/org/mongodb/morphia/utils/IndexType.java
```diff
@@ -11,10 +11,25 @@
...
+        type = o;
+    }
+
+    /**
+     * Returns the enum instance for the given value
+     *
+     * @param value the value to find
+     * @return the enum instance
+     */
+    public static IndexType fromValue(final Object value) {
```
jyemin: unit test

morphia/src/test/java/org/mongodb/morphia/TestIdField.java
```diff
@@ -96,12 +96,12 @@ public void testMapAsId() throws Exception {
     @Entity
```
jyemin: I'm not clear how anything from here down is related to the index changes above.

morphia/src/test/java/org/mongodb/morphia/indexes/TestIndexes.java
```diff
@@ -83,9 +85,8 @@ private void assertHashed(final List<DBObject> indexInfo) {
     }
 
     @Entity(noClassnameStored = true)
-    @Indexes({@Index(options = @IndexOptions(), fields = {@Field(value = "name")})})
+    @Indexes({@Index(options = @IndexOptions(collation = @Collation(locale = "en_US")), fields = {@Field(value = "name")})})
```
jyemin: There should be an assertion that collation was properly added to the index.

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
```diff
@@ -0,0 +1,492 @@
...
+            indexOptions.languageOverride(options.languageOverride());
+        }
+        if (!options.name().equals("")) {
+            indexOptions.name(options.name());
+        }
+        if (options.expireAfterSeconds() != -1) {
+            indexOptions.expireAfter((long) options.expireAfterSeconds(), TimeUnit.SECONDS);
+        }
+        if (!options.collation().locale().equals("")) {
+            indexOptions.collation(convert(options.collation()));
```
jyemin: No tests fail if I comment this line out.  There should be at least one that does.

morphia/src/main/java/org/mongodb/morphia/mapping/MappedClass.java
```diff
@@ -362,7 +364,7 @@ public Annotation getFirstAnnotation(final Class<? extends Annotation> clazz) {
      * @param clazz the annotation to search with
      * @return the list of methods
      */
-    public List<ClassMethodPair> getLifecycleMethods(final Class<Annotation> clazz) {
+    List<ClassMethodPair> getLifecycleMethods(final Class<Annotation> clazz) {
```
jyemin: Why the binary compat break?

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -247,6 +254,16 @@
...
 
     /**
+     * Returns the {@see MongoCollection} for the given name
+     *
+     * @param <T>   the type of the MongoCollection
+     * @param name  the name of the collection
+     * @param clazz the type of the collection
+     * @return the collection
+     */
+    <T> MongoCollection<T> getMongoCollection(String name, Class<T> clazz);
```
evanchooly: IndexHelper gets a `Datastore` reference and needs to get `MongoCollection` instances

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -463,13 +399,30 @@ public void ensureIndexes(final boolean background) {
         return getByKeys(null, keys);
     }
 
+    private DBCollection getCollection(final Object obj) {
+        if (obj == null) {
+            return null;
```
evanchooly: But this is checking that the input is not null.

jyemin: I get that.  But as a private method, is there ever a case where the caller would be expected to sometimes pass in null and then check for a null response?  If not, better to just let this method throw rather than return null.   The caller could just as easily not call the method if the argument is null.

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
jyemin: I see that it's used by the IndexHelper#createIndex methods.  How about the caller (which is just DatastoreImpl anyway) just passes in the MongoCollection (or the MongoDatabase) as an extra parameter to createIndex?  Then IndexHelper doesn't have to call back to Datastore just to get it.

morphia/src/test/java/org/mongodb/morphia/indexes/TestIndexes.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: It gets called from a number of places that accept user input.  There's a decent chance nulls might make it in here.

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: It's no less an abstraction leak than the existing getCollection() methods.  It can probably hidden in *this* case but the plan is to move everything to `MongoCollection` eventually and this will only open back up.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
jyemin: That's only a valid argument if it's really ok for the user input itself to be null.  The one test that exercises this code path is TestDatastore#saveNull, but that's only because DatastoreImpl#save(T, com.mongodb.WriteConcern) doesn't check for a null entity until after calling DatastoreImpl#getCollection(java.lang.Object).

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -1027,17 +1021,6 @@ public void setDecoderFact(final DBDecoderFactory fact) {
     }
 
     /**
-     * @param obj the instance to use for looking up the collection mapping
-     * @return the collection mapped for the type of obj
-     */
-    public DBCollection getCollection(final Object obj) {
```
jyemin: What do you mean?  Moved where?

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
jyemin: Let's leave it for when we execute that plan.  For now keep it out of the public API.

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
```diff
@@ -0,0 +1,492 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static java.lang.String.format;
+import static java.lang.reflect.Proxy.newProxyInstance;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static org.mongodb.morphia.utils.IndexType.fromValue;
+
+class IndexHelper {
```
jyemin: Can you point me to them?

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
jyemin: Acknowledged

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
jyemin: Acknowledged

morphia/src/main/java/org/mongodb/morphia/annotations/IndexOptions.java
```diff
@@ -78,4 +79,9 @@
      * Creates the index as a unique value index; inserting duplicates values in this field will cause errors
      */
     boolean unique() default false;
+
+    /**
+     * Defines the collation to be used for this index
+     */
+    Collation collation() default @Collation(locale = "");
```
jyemin: Acknowledged.  Really annoying that null can't be the default value

morphia/src/main/java/org/mongodb/morphia/mapping/Mapper.java
jyemin: Acknowledged, but please explain how it's related to the collation change.

morphia/src/test/java/org/mongodb/morphia/TestIdField.java
jyemin: Still not sure how the interface resolution work is related to collation.

evanchooly: Trying to translate the key values to reflect the document names vs the java field names brought this issues to the surface.

morphia/src/main/java/org/mongodb/morphia/mapping/Mapper.java
jyemin: If I remove this change all the tests still pass.

I suggest reverting this and handling in a separate Issue/PR.

morphia/src/main/java/org/mongodb/morphia/utils/IndexType.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
evanchooly: that would just mean the index gets created without a collation but that's not an error state.

morphia/src/main/java/org/mongodb/morphia/Datastore.java
```diff
@@ -323,6 +331,14 @@
     DBCollection getCollection(Class<?> clazz);
 
     /**
+     * @param <T> the type of the MongoCollection
+     * @param clazz the class to use for mapping
+     * @return the mapped collection for the collection
+     * @see MongoCollection
+     */
+    <T> MongoCollection<T> getMongoCollection(Class<T> clazz);
```
evanchooly: removed

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: It's (mostly) not ok for user input to be null.  but this is the first time it gets checked on any of the paths to this point.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: up above where we're debating the null check

morphia/src/main/java/org/mongodb/morphia/mapping/Mapper.java
evanchooly: if you comment out the interface check and the `@Embededed` on `Nested` in `EmbeddedMappingTest` then run `org.mongodb.morphia.TestMapper#subTypes` you'll see the failure.

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
jyemin: Woops, my comment was misleading.  I meant that if I comment out lines 101-103, no tests fail, which means there are no tests that assert that the create index contains the correct collation.

This makes me question whether the rest of IndexHelper is sufficiently covered by tests.

morphia/src/main/java/org/mongodb/morphia/mapping/Mapper.java
jyemin: Acknowledged, but still it is orthogonal to collation as far as I know.  So unless I'm wrong about that it should be handled separately.

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
evanchooly: https://github.com/mongodb/morphia/blob/7bd6a89067dcca0d3ecccf99668ec8f3e9da5827/morphia/src/test/java/org/mongodb/morphia/indexes/TestIndexes.java#L65-L65

This test looked for the locale setting.  I went ahead and expanded it to check all the other values as well.

morphia/src/main/java/org/mongodb/morphia/IndexHelper.java
evanchooly: TestIndexes
TestIndexed
TestIndexCollections
TestIndexInheritance
TestEmbeddedArrayIndexes
TestTextIndexing

evanchooly:
ready for another review

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
jyemin: Since this method has to made public again anyway to preserve binary compat, best to leave the behavior as is since applications may be relying on it.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -78,49 +70,29 @@
...
 
-    /**
-     * Create a new DatastoreImpl
-     *
-     * @param morphia     the Morphia instance
-     * @param mongoClient the connection to the MongoDB instance
-     * @param dbName      the name of the database for this data store.
-     */
-    public DatastoreImpl(final Morphia morphia, final MongoClient mongoClient, final String dbName) {
+    DatastoreImpl(final Morphia morphia, final MongoClient mongoClient, final String dbName) {
```
jyemin: So you can deprecate the DatastoreImpl class, but you still shouldn't change any existing constructors.  So can you put this back as is?

morphia/src/main/java/org/mongodb/morphia/annotations/IndexOptions.java
```diff
@@ -45,8 +45,9 @@
     boolean disableValidation() default false;
 
     /**
-     * Tells the unique index to drop duplicates silently when creating; only the first will be kept
+     * @deprecated This value has been desupported.  Setting this to true will result in an error.
```
jyemin: Not seeing the fix.

evanchooly: done

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java

evanchooly:
PR approved.  will merge to master once the "operations" branch is merged to this one.

PR #1034, opened 2016-10-05
Collation support in operations

evanchooly:
This PR adds support for collations to the various operations on `Query` and `Datastore`.  As a function of how the driver accepts options, this PR also exposes new Options classes to pass along to the driver.  The API for `Query` and `Datastore` have been updated to take these new options and the old methods that take a varying number of parameters have been deprecated in favor of the new Options based methods.


jyemin:
I thought a bit more about my previous comment, and it's not really apt.  Given the way collation is being exposed, it _was_ necessary to do it at the same time as the new options classes.  

What actually made the review challenging is the number of changes to the implementation that either aren't strictly required by the addition of options classes, or else it wasn't clear why it was required.   When making a change as large as this, it's useful to change as little else as possible at the same time, or else explain exactly why the additional changes were necessary.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -1316,22 +1316,39 @@ private long nextValue(final Long oldVersion) {
...
+            LOG.trace(format("Executing update(%s) for query: %s, ops: %s, multi: %s, upsert: %s",
+                             dbColl.getName(), queryObject, update, options.isMulti(), options.isUpsert()));
+        }
+
+        return new UpdateResults(dbColl.update(queryObject, update, options.getOptions()));
     }
 
+
     @SuppressWarnings("unchecked")
     private <T> UpdateResults update(final Query<T> query, final DBObject update, final boolean createIfMissing, final boolean multi,
```
jyemin: Acknowledged

morphia/src/main/java/org/mongodb/morphia/query/CountOptions.java
```diff
@@ -0,0 +1,261 @@
...
+            && (getHint() == null && that.getHint() == null || getHint().equals(that.getHint()))
+            && (getReadPreference() == null && that.getReadPreference() == null || getReadPreference().equals(that.getReadPreference()))
+            && (getReadConcern() == null && that.getReadConcern() == null || getReadConcern().equals(that.getReadConcern()))
+            && (getCollation() == null && that.getCollation() == null || getCollation().equals(that.getCollation()));
+
+    }
+
+    @Override
+    public int hashCode() {
+        return getOptions().hashCode();
```
jyemin: Even though equals is for testing, if you're going to have equals and hashCode in a public class, they shouldn't violate the equals-consistent-with-hash-code rule.  So this still needs to be addressed.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -286,30 +285,21 @@ public void ensureCaps() {
...
+                                           .remove(true);
 
-        if (result != null) {
-            return mapper.fromDBObject(this, query.getEntityClass(), result, createCache());
-        }
+        final DBObject result = dbColl.findAndModify(query.getQueryObject(), copy.getOptions());
 
-        return null;
-    }
+        return mapper.fromDBObject(this, query.getEntityClass(), result, createCache());
```
jyemin: Still, this should be rolled back, because Mapper.fromDBObject logs an error when passed a null dbObject and it's not an error in this situation.

morphia/src/main/java/org/mongodb/morphia/query/QueryImpl.java
```diff
@@ -45,26 +48,16 @@
...
-    private ReadPreference readPref;
-    private Integer maxScan;
-    private Long maxTime;
-    private TimeUnit maxTimeUnit;
-    private String comment;
-    private boolean returnKey;
+    private FindOptions options;
+
+    FindOptions getOptions() {
+        if (options == null) {
```
jyemin: My concern is that it's error-prone, as it requires all internal users of options to get it through this method, and it would be easy to forget to do that.  

But it's up to you.  If you don't think it's a problem, then leave as is.

morphia/src/main/java/org/mongodb/morphia/query/QueryImpl.java
```diff
@@ -279,14 +310,17 @@ protected BasicDBObject copy(final BasicDBObject dbObject) {
...
+    @Deprecated
     public Query<T> disableCursorTimeout() {
-        noTimeout = true;
+        getOptions().noCursorTimeout(true);
         return this;
     }
 
     @Override
+    @Deprecated
     public Query<T> disableSnapshotMode() {
```
jyemin: You could unit test it at least.

jyemin: Also, BSONObject has a removeField method, so you don't need to get it as a BasicDBObject to remove the $snapshot field.

morphia/src/main/java/org/mongodb/morphia/query/QueryImpl.java
```diff
@@ -558,8 +616,9 @@ private void validateProjections(final String field, final boolean include) {
     }
 
     @Override
+    @Deprecated
     public Query<T> returnKey() {
```
jyemin: Acknowledged

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -1316,22 +1316,39 @@ private long nextValue(final Long oldVersion) {
...
-        if (((UpdateOpsImpl) ops).isIsolated()) {
-            final Query<T> q = query.cloneQuery();
-            q.disableValidation().filter("$atomic", true);
-            return update(q, u, createIfMissing, multi, wc);
+        final MappedClass mc = getMapper().getMappedClass(query.getEntityClass());
+        final List<MappedField> fields = mc.getFieldsAnnotatedWith(Version.class);
+
+        DBObject queryObject = query.getQueryObject();
+        if (operations.isIsolated()) {
+            queryObject.put("$isolated", true);
```
jyemin: Line 1332 is also not covered bty any TestDatastore tests.  A test should be added at the least.

morphia/src/main/java/org/mongodb/morphia/query/QueryImpl.java
evanchooly: the getBlahDBObject() methods were there so that I could migrate to `Document` based code at some point.  Instead, i've just removed them from the public API altogether so I can change them in the future without breaking user code.

jyemin:
LGTM

PR #1037, opened 2016-10-10
Indexhelper tests

evanchooly:
This adds a number of explicit tests for various indexing definitions.  It involves some refactoring to expose certain methods and replaces the `synthesize*` methods for creating annotations on the fly with different `Builder` implementations.


rozza:
Lots of deprecation warnings and a couple of failed tests in travis.  Want to fix them before I review?

evanchooly:
The deprecations i'm going to pull out in to a separate branch/commit so you don't need to worry about those.  All the tests pass locally and I don't see on travis which ones are failing.

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
```diff
@@ -0,0 +1,443 @@
...
+    public void findField() {
+        MappedClass mappedClass = getMorphia().getMapper().getMappedClass(IndexedClass.class);
+
+        assertEquals("name", indexHelper.findField(mappedClass, new IndexOptionsBuilder(), singletonList("indexName")));
+        assertEquals("name", indexHelper.findField(mappedClass, new IndexOptionsBuilder(), singletonList("name")));
+        assertEquals("nest.name", indexHelper.findField(mappedClass, new IndexOptionsBuilder(), asList("nested", "name")));
+        assertEquals("nest.name", indexHelper.findField(mappedClass, new IndexOptionsBuilder(), asList("nest", "name")));
+
+        try {
+            assertEquals("nest.whatsit", indexHelper.findField(mappedClass, new IndexOptionsBuilder(), asList("nest", "whatsit")));
```
rozza: Nit: Do you need the assertEquals here if the lookup will fail?

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
```diff
@@ -0,0 +1,443 @@
...
+            fail("Should have failed on the bad index path");
+        } catch (MappingException e) {
+            // alles ist gut
+        }
+        assertEquals("nest.whatsit.nested.more.deeply.than.the.object.model",
```
rozza: Is it the disable validation flag that allows this one to pass rather than the previous example?

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
```diff
@@ -0,0 +1,443 @@
...
+        assertEquals("index_name", options.getName());
+        assertTrue(options.isBackground());
+        assertTrue(options.isUnique());
+        assertTrue(options.isSparse());
+        assertEquals(Long.valueOf(42), options.getExpireAfter(TimeUnit.SECONDS));
+        assertEquals("en", options.getDefaultLanguage());
+        assertEquals("de", options.getLanguageOverride());
+        assertEquals(indexHelper.convert(indexOptions.collation()), options.getCollation());
+
+        assertTrue(indexHelper.convert(indexOptions, true).isBackground());
```
rozza: This background conversion logic may warrant a test case for itself

morphia/src/main/java/org/mongodb/morphia/AnnotationBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+import org.mongodb.morphia.mapping.MappingException;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.lang.String.format;
+
+abstract class AnnotationBuilder<T extends Annotation> implements Annotation {
```
rozza: Should I be reviewing this as? Not sure how it fits in with the IndexHelper tests?

morphia/src/main/java/org/mongodb/morphia/annotations/Indexed.java
```diff
@@ -76,10 +76,7 @@
     boolean unique() default false;
 
     /**
-     * Indicates the direction of the index (ascending, descending); default is ascending
-     *
-     * @deprecated use the {@link IndexOptions} found in {@link #options()}
+     * Indicates the type of the index (ascending, descending); default is ascending
      */
-    @Deprecated
```
evanchooly: because the other values are options to the index and this isn't.  i.e., there isn't a corresponding value in IndexOptions for this.  at the class level, this would defined in the @Field annotation.

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
evanchooly: if findField() fails that's a different kind of failure than properly mapping path segments to their document names

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
evanchooly: yes

morphia/src/main/java/org/mongodb/morphia/AnnotationBuilder.java
evanchooly: It's what I'm using to create runtime annotation instances.  a logic check certainly wouldn't hurt.

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
evanchooly: There is one.  org.mongodb.morphia.IndexHelperTest#indexOptionsConversion

morphia/src/main/java/org/mongodb/morphia/annotations/Indexed.java
rozza: Acknowledged, was it released as deprecated then?  What did users move to?  Or was it a mid release change and change back?  Either way, I'm happy that you're happy with it! :)

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
rozza: So is all that needed?

evanchooly: it's just testing that values copy over correctly.  My big complaint is that we'll have to manually track evolutions of the object manually.  I'm pondering some sort of reflection based test so that methods are discovered and tested automatically.  That way we don't have to remember to update the tests when new values show up.  But that's for another time.

morphia/src/main/java/org/mongodb/morphia/annotations/Indexed.java
evanchooly: It probably was released as deprecated.  This particular annotation isn't used as much so it doesn't seem as if anyone ran in to it.

morphia/src/main/java/org/mongodb/morphia/AnnotationBuilder.java
```diff
@@ -0,0 +1,104 @@
...
+                values.put(method.getName(), method.invoke(original));
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    <V> V get(final String key) {
+        return (V) values.get(key);
```
rozza: This is a bit magical to me - I've seen similar in Scala but not in Java.  Does the compiler fill in what type V is?

morphia/src/main/java/org/mongodb/morphia/CollationBuilder.java
```diff
@@ -0,0 +1,120 @@
...
+
+class CollationBuilder extends AnnotationBuilder<Collation> implements Collation {
+    @Override
+    public Class<Collation> annotationType() {
+        return Collation.class;
+    }
+
+    @Override
+    public boolean backwards() {
+        return get("backwards");
```
rozza: I didn't know the compiler could be used to set the result type for the get method. Pure voodoo!

morphia/src/main/java/org/mongodb/morphia/IndexBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+    public boolean unique() {
+        return get("unique");
+    }
+
+    @Override
+    public String value() {
+        return get("value");
+    }
+
+    private List<Field> parseFieldsString(final String str) {
```
rozza: Thats a bit of a scary method.  Is the string something the user will define? Does there need to be any error checking / validation?

morphia/src/main/java/org/mongodb/morphia/IndexOptionsBuilder.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+package org.mongodb.morphia;
+
+import org.mongodb.morphia.annotations.Collation;
+import org.mongodb.morphia.annotations.Index;
+import org.mongodb.morphia.annotations.IndexOptions;
+import org.mongodb.morphia.annotations.Indexed;
+
+class IndexOptionsBuilder extends AnnotationBuilder<IndexOptions> implements IndexOptions {
```
rozza: Is this needed for legacy reasons? I'm trying to understand its role compared to the `IndexBuilder` and `IndexedBuilder`

morphia/src/main/java/org/mongodb/morphia/AnnotationBuilder.java
evanchooly: V is inferred from the context of the call site.

morphia/src/main/java/org/mongodb/morphia/IndexBuilder.java
evanchooly: the field names eventually get validated against the model.  But yes, it's scary, and that's why this "feature" is being deprecated.  :D

morphia/src/main/java/org/mongodb/morphia/IndexOptionsBuilder.java
evanchooly: @Index applies to classes and @Indexed applies directly to fields.  Other than that, i'm not sure i understand what you're asking.

evanchooly:
merged

PR #1039, opened 2016-10-11
Doc validation

evanchooly:
Opening a new since #981 went sideways.  The conversation is still there, at least.


morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -256,6 +263,41 @@ public void ensureCaps() {
...
+
+            if (!result.ok()) {
+                if (result.getInt("code") == 26) {
+                    try {
+                        ValidationOptions options = new ValidationOptions()
+                            .validator(parse(validation.value()))
+                            .validationLevel(validation.level())
+                            .validationAction(validation.action());
+                        getDatabase().createCollection(collectionName, new CreateCollectionOptions().validationOptions(options));
+                    } catch (JsonParseException e) {
```
jyemin: This is inconsistent with the collMod case above, where JsonParseException is not caught.  I think it's better not to catch it as it's a real error that the user should deal with.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -256,6 +263,41 @@ public void ensureCaps() {
...
+                        LOG.warning(format("Could not parse validator for '%s:'  %s", collectionName, e.getMessage()));
+                    }
+                } else {
+                    LOG.warning(format("Could not add document validation on '%s:'  %s", collectionName, result.getErrorMessage()));
```
jyemin: This is going to create inconsistent results.  If collMod fails for some real reason (like not authorized to modify the collection), it will just log a warning.  But if create fails for the same reason, it will throw an exception.

What does index creation do if there's an error creating the index?  I suggest we make the behavior here consistent with whatever that is.

morphia/src/main/java/org/mongodb/morphia/annotations/Validation.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import static com.mongodb.client.model.ValidationAction.ERROR;
+import static com.mongodb.client.model.ValidationLevel.STRICT;
+
+
+/**
+ * Defines the document validation logic for a collection.
```
jyemin: Add a comment that the driver does not attempt to replace field names, as it does with indexes and queries.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: The exception percolates up.  I'll change this from logging to throwing an exception.

morphia/src/main/java/org/mongodb/morphia/annotations/Validation.java
evanchooly: done.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -256,6 +262,38 @@ public void ensureCaps() {
...
+
+            if (!result.ok()) {
+                if (result.getInt("code") == 26) {
+                    ValidationOptions options = new ValidationOptions()
+                        .validator(parse(validation.value()))
+                        .validationLevel(validation.level())
+                        .validationAction(validation.action());
+                    getDatabase().createCollection(collectionName, new CreateCollectionOptions().validationOptions(options));
+                } else {
+                    throw new MappingException(format("Could not add document validation on '%s:'  %s", collectionName,
```
jyemin: Not sure what you do in other places, but result.throwOnError() will throw an exception that includes pertinent information like the error code.

evanchooly: i'll update to that.  I'll add that in place of the if !ok check.

jyemin:
lgtm barring assuming Travis is green

evanchooly:
merged

PR #1042, opened 2016-10-14
deprecations and documentation

evanchooly:
The introduction of the options classes allows for a consolidation of overloads on a few classes.  This PR deprecates the obvious candidates and updates the documentation to reflect the new preferred methods.  In the cases where there isn't a helper method  (e.g., $comment, $snapshot) available, the javadoc has been updated to reflect how to achieve the desired functionality with the new API.


evanchooly:
any questions/objections?

jyemin:
I have no questions, but I do have a concern that this won't be received well by the community.  

So LGTM but let's see how it's received in a release candidate.

evanchooly:
I'm a bit curious about that as well.  I think it's the right decision but we'll see how it goes.

Issue #735, opened 2015-03-20
Query.order(String) API design, to overwrite values set in previous calls, is flawed


ineedteepee:
The implementation of Query.order(String) overwrites the values set in previous calls to order(String) (an undocumented feature). The following statement should have expected results:

query.order("name").order("-time");

Issue #730 was closed with the blanket answer "works as designed and documented". While it works as documented (yes, technically it can do anything and still work as documented since there is no documentation about it) it does not work _as expected_.

The purpose of fluent interfaces is to reduce developer burden. The current design requires the programmer to build a comma separated string in order to set a sort order to query. This is not a problem if the developer is working with a static set of parameters but if the set of ordering properties is dynamic, as is my use case, this fluent API is very cumbersome. This is not how fluent interfaces should work. The order method even returns a Query object which invites the developer to call the order method again for the same object.

A query can have several filters. These filters are added to the query with the filter(String,Object) method.
query.filter("a", 1).filter("b", 2);

Respectively a query can have several ordering properties. The API is not consistent here.

As I said in #730. I have the fix for this ready with unit tests and all. If the API design is an ego issue, just say so and I won't try to contribute to your project again. Otherwise, please address the arguments I have made.

evanchooly:
1.  It's not an ego issue.  It's existing behavior that's been around for years.
2.  It *is* documented quite well:  https://rawgit.com/wiki/mongodb/morphia/javadoc/0.110/org/mongodb/morphia/query/Query.html#order-java.lang.String-

Contributions are certainly welcome but changing existing behaviors isn't something that should be done lightly.  Personally, I'd have probably written something like what you suggest had I written that interface.  But it's been working as it is for a few years now with no complaints.  Changing it now introduces a certain level of risk I'm not comfortable with when I'm trying to lock down a final, long overdue 1.0 release.

That said, patches trump bug requests.  If we were to introduce a new set of methods (e.g., `sortAsc()` and `sortDesc()`) and perhaps deprecate `order()`, that's a bit more palatable. (The C# driver returns a whole new interface from `sort()` to provide stronger API checks, e.g.)  I'm trying to minimize API changes at this point, though, because I really, really want a formal 1.0 soon so that we *can* start morphing the API in a 2.0 effort.  So show us your pull request and we can talk suitability and then maybe timing.

evanchooly:
explore updating API to use the `Sort` type from the `AggregationPipeline` API.

Issue #499, opened 2013-07-30
add natural order support


evanchooly:
http://docs.mongodb.org/manual/reference/operator/natural/#op._S_natural

cstrempfer:
I could add an solution for that to pull request https://github.com/mongodb/morphia/pull/1043. Should there be explicit methods for that, e.g. `Sort#ascendingNatural()`, or is adding a constant `NATURAL` to the `Sort` class enough? I prefer the first option, because the constant could easily be missed by developers.

evanchooly:
I favor the explicit methods, too, though I wonder if `naturalAscending()` and `naturalDescending()` might be better.  Either way is probably, fine, though.  As for the package issue, I think a new `Sort` should be added outside the `aggregation` package.  I'd use `order` rather than `direction`.  The aggregation package's Sort would subclass the new one and cast `order` to `int` to preserve the existing API.  Does that make sense?

cstrempfer:
> I'd use `order` rather than `direction`.

Do you talk about the package name? If yes, I move it to `order`. If not, is the `utils` package, ok?

> The aggregation package's Sort would subclass the new one and cast `order` to `int` to preserve the existing API.

What's the `order` part? I cannot find an `order` class in the master branch. Do you mean creating a new enum for the direction?

PR #1043, opened 2016-10-14
use Sort vararg for ordering of query results

cstrempfer:
Instead of
`datastore.find(Rectangle.class).order("width,-height").get();`
you can now write
`datastore.find(Rectangle.class).order(ascending("width"), descending("height")).get();`

This partially solves https://github.com/mongodb/morphia/issues/735. I didn't implement methods to chain sort parameters.

The `Sort` class is in `org.mongodb.morphia.aggregation` right now. It could be moved to `org.mongodb.morphia.utils` as it's used for queries and aggregation now, but that will break existing code.


evanchooly:
Thanks!  This looks great.  I have a few tweaks I want to make to address the package issue you mentioned but I think we can get this merged next week.

cstrempfer:
@evanchooly I've made the changes you suggested in #499. It wasn't necessary to change the type of `Sort#order` (aka. `Sort#direction`).

I didn't change the unit tests of sorting in an aggregation pipeline to make sure the old API still works.

PR #1048, opened 2016-10-18
add support for partial filter indexes

evanchooly:


morphia/src/main/java/org/mongodb/morphia/annotations/IndexOptions.java
```diff
@@ -86,5 +86,11 @@
      * Defines the collation to be used for this index
      * @since 1.3
      */
+    String partialFilter() default "";
```
jyemin: Is there a good place to put an integration test (where partial filter is specified on an entity with an annotation), perhaps in the org.mongodb.morphia.indexes test package?

jyemin:
LGTM

PR #1049, opened 2016-10-18
add tests for bypass document validation

### LINKED ISSUE #1023: expand bypassDocumentValidation coverage
Some places like aggregation already have coverage due to usage of the AggregationOptions class from the driver.  Double check the other areas of usage to ensure Morphia exposes the opportunity set this option.

evanchooly:
fixes #1023

(restoring a mistaken branch deletion)

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
+++ b/morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
@@ -257,16 +257,6 @@ public interface AdvancedDatastore extends Datastore {
     void setDecoderFact(DBDecoderFactory fact);
 
     /**
-     * Inserts an entity in to the named collection.
-     *
-     * @param collection the collection to update
-     * @param entity     the entity to insert
-     * @param <T>        the type of the entity
-     * @return the new key of the inserted entity
-     */
-    <T> Key<T> insert(String collection, T entity);
-
-    /**
      * Inserts an entity in to the mapped collection.
      *
      * @param entity the entity to insert
@@ -279,21 +269,56 @@ public interface AdvancedDatastore extends Datastore {
      * Inserts an entity in to the mapped collection.
      *
      * @param entity the entity to insert
-     * @param <T>    the type of the entity
-     * @return the new key of the inserted entity
-     */
-    <T> Iterable<Key<T>> insert(Iterable<T> entity);
-
-    /**
-     * Inserts an entity in to the mapped collection.
-     *
-     * @param entity the entity to insert
      * @param wc     the WriteConcern to use when inserting
      * @param <T>    the type of the entity
      * @return the new key of the inserted entity
      * @see WriteConcern
+     * @deprecated use {@link #insert(Object, InsertOptions)}
      */
+    @Deprecated
     <T> Key<T> insert(T entity, WriteConcern wc);
+
+    /**
+     * Inserts an entity in to the mapped collection.
+     *
+     * @param entity  the entity to insert
+     * @param options the options to apply to the save operation
+     * @param <T>     the type of the entity
+     * @return the new key of the inserted entity
+     * @since 1.3
+     */
+    <T> Key<T> insert(T entity, InsertOptions options);
+
+    /**
+     * Inserts an entity in to the named collection.
+     *
+     * @param collection the collection to update
+     * @param entity     the entity to insert
+     * @param <T>        the type of the entity
+     * @return the new key of the inserted entity
+     */
+    <T> Key<T> insert(String collection, T entity);
+
+    /**
+     * Inserts an entity in to the named collection.
+     *
+     * @param collection the collection to update
+     * @param entity     the entity to insert
+     * @param options    the options to apply to the save operation
+     * @param <T>        the type of the entity
+     * @return the new key of the inserted entity
+     * @since 1.3
+     */
+    <T> Key<T> insert(String collection, T entity, InsertOptions options);
+
+    /**
+     * Inserts entities in to the mapped collection.
+     *
+     * @param entities the entities to insert
+     * @param <T>    the type of the entities
+     * @return the new key of the inserted entities
+     */
+    <T> Iterable<Key<T>> insert(Iterable<T> entities);
 
     /**
      * Inserts entities in to the mapped collection.
@@ -301,7 +326,9 @@ public interface AdvancedDatastore extends Datastore {
      * @param entities the entities to insert
      * @param <T>      the type of the entity
      * @return the new keys of the inserted entities
+     * @deprecated use {@link #insert(Iterable)} instead
      */
+    @Deprecated
     <T> Iterable<Key<T>> insert(T... entities);
 
     /**
@@ -311,8 +338,21 @@ public interface AdvancedDatastore extends Datastore {
      * @param wc       the WriteConcern to use when inserting
      * @param <T>      the type of the entity
      * @return the new keys of the inserted entities
+     * @deprecated use {@link #insert(Iterable, InsertOptions)}
      */
+    @Deprecated
     <T> Iterable<Key<T>> insert(Iterable<T> entities, WriteConcern wc);
+
+    /**
+     * Inserts entities in to the mapped collection.
+     *
+     * @param entities the entities to insert
+     * @param options  the options to apply to the save operation
+     * @param <T>      the type of the entity
+     * @return the new keys of the inserted entities
+     * @since 1.3
+     */
+    <T> Iterable<Key<T>> insert(Iterable<T> entities, InsertOptions options);
 
     /**
      * Inserts an entity in to the named collection.
@@ -334,8 +374,22 @@ public interface AdvancedDatastore extends Datastore {
      * @param <T>        the type of the entity
      * @return the new keys of the inserted entities
      * @see WriteConcern
+     * @deprecated use {@link #insert(String, Iterable, InsertOptions)} instead
      */
+    @Deprecated
     <T> Iterable<Key<T>> insert(String collection, Iterable<T> entities, WriteConcern wc);
+
+    /**
+     * Inserts an entity in to the named collection.
+     *
+     * @param collection the collection to update
+     * @param entities   the entities to insert
+     * @param options    the options to apply to the save operation
+     * @param <T>        the type of the entity
+     * @return the new keys of the inserted entities
+     * @since 1.3
+     */
+    <T> Iterable<Key<T>> insert(String collection, Iterable<T> entities, InsertOptions options);
 
     /**
      * Returns a new query based on the example object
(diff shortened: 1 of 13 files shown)


evanchooly:
That's a good catch.  I'm pretty sure it'll need something, too.

evanchooly:
Two things:

1. yes, it needs a test for this
1. The API is broken enough to make this difficult.  I'd like to defer this to the 1.4 release.

jyemin:
Please elaborate.  What exactly is too broken to fix now?  I'd rather not partially support bypassDocumentValidation if we can help it.

evanchooly:
Actually, i just thought of a way to test this with the current API.  stand by...

jyemin:
Aggregate with $out looks good.  How about insert?

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -257,51 +257,78 @@
...
+    @Deprecated
+    <T> Key<T> insert(T entity, WriteConcern wc);
 
     /**
      * Inserts an entity in to the mapped collection.
      *
-     * @param entity the entity to insert
-     * @param <T>    the type of the entity
+     * @param entity  the entity to insert
+     * @param options the options to apply to the save operation
```
jyemin: nit: to the insert operation

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -257,51 +257,78 @@
...
      */
-    <T> Key<T> insert(T entity, WriteConcern wc);
+    <T> Key<T> insert(String collection, T entity);
+
+    /**
+     * Inserts an entity in to the named collection.
+     *
+     * @param collection the collection to update
+     * @param entity     the entity to insert
+     * @param options    the options to apply to the save operation
```
jyemin: nit: to the insert operation

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -257,51 +257,78 @@
...
+     * @since 1.3
+     */
+    <T> Key<T> insert(String collection, T entity, InsertOptions options);
+
+    /**
+     * Inserts entities in to the mapped collection.
+     *
+     * @param entities the entities to insert
+     * @param <T>    the type of the entities
+     * @return the new key of the inserted entities
```
jyemin: nit: the new keys

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -311,10 +338,23 @@
...
+     * @deprecated use {@link #insert(Iterable, InsertOptions)}
      */
+    @Deprecated
     <T> Iterable<Key<T>> insert(Iterable<T> entities, WriteConcern wc);
 
     /**
+     * Inserts entities in to the mapped collection.
+     *
+     * @param entities the entities to insert
+     * @param options  the options to apply to the save operation
```
jyemin: nit: to the insert operation

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -334,10 +374,24 @@
      * @param <T>        the type of the entity
      * @return the new keys of the inserted entities
      * @see WriteConcern
+     * @deprecated use {@link #insert(String, Iterable, InsertOptions)} instead
      */
+    @Deprecated
     <T> Iterable<Key<T>> insert(String collection, Iterable<T> entities, WriteConcern wc);
 
     /**
+     * Inserts an entity in to the named collection.
```
jyemin: Nit: Inserts entities

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -334,10 +374,24 @@
...
+     *
+     * @param collection the collection to update
+     * @param entities   the entities to insert
+     * @param options    the options to apply to the save operation
```
jyemin: nit: to the insert operation

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -365,7 +419,20 @@
...
      */
+    @Deprecated
     <T> Key<T> save(String collection, T entity, WriteConcern wc);
 
+    /**
+     * Saves an entity in to the named collection.
+     *
+     * @param collection the collection to update
+     * @param entity     the entity to insert
+     * @param <T>        the type of the entity
```
jyemin: nit: consistent ordering (param <T> last)

morphia/src/main/java/org/mongodb/morphia/Datastore.java
```diff
@@ -575,16 +575,30 @@
      * @param <T>      the type of the entity
      * @param wc       the WriteConcern to use
      * @return the keys of the entities
+     * @deprecated use {@link #save(Iterable, InsertOptions)} instead
      */
+    @Deprecated
     <T> Iterable<Key<T>> save(Iterable<T> entities, WriteConcern wc);
 
     /**
+     * Saves the entities (Objects) and updates the @Id field, with the WriteConcern
```
jyemin: Why does this specifically call out the update to the @Id field?   Does this save method have different behavior than the one in Datastore?  I would expect all save, and insert methods, to update the @Id-annotated field.  

Let's use consistent language regarding update to @Id.  Either call it out everywhere it happens, or nowhere.

morphia/src/main/java/org/mongodb/morphia/Datastore.java
```diff
@@ -575,16 +575,30 @@
...
+     *
+     * @param entities the entities to save
+     * @param <T>      the type of the entity
```
jyemin: nit: move down

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -702,44 +703,50 @@ public void setQueryFactory(final QueryFactory queryFactory) {
...
 
     @Override
+    @Deprecated
     public <T> Iterable<Key<T>> save(final T... entities) {
-        final List<Key<T>> savedKeys = new ArrayList<Key<T>>();
-        for (final T ent : entities) {
-            savedKeys.add(save(ent));
-        }
-        return savedKeys;
+        return save(asList(entities), new InsertOptions());
```
jyemin: This looks like a behavioral change.  In the previous version, each entity is saved with the default write concern for the entity by calling DatastoreImpl#getWriteConcern.  In this version the default write concern (null) of InsertOptions will be used.

I suggest auditing all usage of InsertOptions to ensure that the default write concern is always applied.

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -257,51 +257,78 @@
...
-     * @param entity the entity to insert
-     * @param <T>    the type of the entity
+     * @param entity  the entity to insert
+     * @param options the options to apply to the save operation
+     * @param <T>     the type of the entity
      * @return the new key of the inserted entity
+     * @since 1.3
      */
-    <T> Iterable<Key<T>> insert(Iterable<T> entity);
+    <T> Key<T> insert(T entity, InsertOptions options);
```
jyemin: What does it/should it mean to pass as an argument an InsertOptions instance with a null writeConcern property?  The Java driver, in similar situations, applies the default write concern for the DBCollection.  I suggest that Morphia do the same by always applying DatastoreImpl#getWriteConcern to any operation that passes an options class instance with a null write concern.

evanchooly: It looks them up when it can:  https://github.com/mongodb/morphia/blob/b94969f8eb6bd61592275a4f708ae580091cb543/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java#L1261

There were a handful of gaps I found while auditing the other operations and I've filled them all in.

morphia/src/main/java/org/mongodb/morphia/Datastore.java
evanchooly: It's just mirroring what the options-less method docs say.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
evanchooly: done

morphia/src/main/java/org/mongodb/morphia/Datastore.java
evanchooly: done

evanchooly:
I'll merge then if there are no other comments.

jyemin:
I still need to review the latest patches.

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -702,44 +703,49 @@ public void setQueryFactory(final QueryFactory queryFactory) {
...
 
     @Override
     public <T> Iterable<Key<T>> save(final Iterable<T> entities, final WriteConcern wc) {
+        return save(entities, new InsertOptions().writeConcern(wc));
+    }
+
+    @Override
+    public <T> Iterable<Key<T>> save(final Iterable<T> entities, final InsertOptions options) {
         final List<Key<T>> savedKeys = new ArrayList<Key<T>>();
         for (final T ent : entities) {
```
jyemin: Just a note to consider for another issue: save could be implemented as a single round trip using bulk write operations

morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -373,7 +373,7 @@
     <T> Iterable<Key<T>> insert(String collection, Iterable<T> entities, WriteConcern wc);
 
     /**
-     * Inserts an entity in to the named collection.
+     * Inserts a entities in to the named collection.
```
jyemin: Nit: this is correct in the original

morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
evanchooly: I was actually  thinking about that in terms of resolving #1052.  This could be converted to just *always* use the bulk write operation and remove that flag altogether.  Is that what you're thinking, too?

jyemin: That's a separate issue.  The issue here is that with bulk update capabilities you can even do a save of multiple entities as a single operation even when the save becomes an upsert due to the presence of an _id field.

PR #1050, opened 2016-10-18
update findAndModify to honor UpdateOperations#isolated

### LINKED ISSUE #406: UpdateOperations with isolated() do not honor DatastoreImpl update with createIfMissing behavior [moved]
This is [Issue 406](http://code.google.com/p/morphia/issues/detail?id=406) moved from a Google Code project.
Added by 2012-04-30T19:00:41.000Z by [jaronsampson](http://code.google.com/u/jaronsampson/).
Please review that bug for more context and additional comments, but update this bug.

Original labels: Type-Defect, Priority-Low

### Original description

    morphia 0.99 / morphia logging 0.99 / morphia validation 0.99
    mongo-java-driver-2.7.3 / mongodb-osx-x86_64-2.0.4
    
The expected upsert behavior using update(query, ops, createIfMissing=true) fails to work when isolated() is used in the UpdateOperations. Instead the query always updates existing record after creating an initial record. Removing isolated() allows this method to be used with the expected upsert behavior. Also, using findAndModify(query, ops, false, true) upserts correctly with isolated() left in the ops.
    
The main difference I can see involves a block in DatastoreImpl, where it is calling disableValidation on the query used in the update. The combination of isolated and update seems to result in the query matching all existing documents in the collection and updating that result set. I believe I verified this by adding a second document to the collection via the CLI, and running the morphia update method again: all documents in the collection ended up with the update calls set operations from my morphia code.
    
            private <T> UpdateResults<T> update(Query<T> query, UpdateOperations ops, boolean createIfMissing, boolean multi, WriteConcern wc) {
                    DBObject u = ((UpdateOpsImpl) ops).getOps();
                    if (((UpdateOpsImpl) ops).isIsolated()) {
                            Query<T> q = query.clone();
                            q.disableValidation().filter("$atomic", true);
                            return update(q, u, createIfMissing, multi, wc);                
                    }
                    return update(query, u, createIfMissing, multi, wc);            
            }
    
compared to this block:
    
            public <T> T findAndModify(Query<T> query, UpdateOperations<T> ops, boolean oldVersion, boolean createIfMissing) {
                    QueryImpl<T> qi = (QueryImpl<T>) query;
    
                    DBCollection dbColl = qi.getCollection();
                    //TODO remove this after testing.
                    if(dbColl == null)
                            dbColl = getCollection(qi.getEntityClass());
    
                    if (log.isTraceEnabled())
                            log.info("Executing findAndModify(&quot; + dbColl.getName() + ") with update &quot;);
    
                    DBObject res = dbColl.findAndModify(qi.getQueryObject(),
                               qi.getFieldsObject(), qi.getSortObject(), false,
                               ((UpdateOpsImpl<T>) ops).getOps(), !oldVersion, createIfMissing);
                    
                    if (res == null) 
                            return null;
                    else
                            return (T) mapr.fromDBObject(qi.getEntityClass(), res, createCache());
            }
    
...which sends the UpdateOperations from UpdateOpsImpl settings up to the driver without changing the QueryImpl fields through disableValidation().
    
Note: I have not stress tested this to see if either method actually isolates the update operation or not.
    
Thanks,
Jaron Sampson

evanchooly:
fixes #406

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
@@ -344,11 +344,16 @@ public class DatastoreImpl implements AdvancedDatastore {
         }
 
         updateForVersioning(query, operations);
-        DBObject res = dbColl.findAndModify(query.getQueryObject(), options.copy()
-                                                                           .sort(query.getSortObject())
-                                                                           .projection(query.getFieldsObject())
-                                                                           .update(((UpdateOpsImpl<T>) operations).getOps())
-                                           .getOptions());
+        DBObject queryObject = query.getQueryObject();
+        if (operations.isIsolated()) {
+            queryObject.put("$isolated", true);
+        }
+
+        DBObject res = dbColl.findAndModify(queryObject, options.copy()
+                                                                .sort(query.getSortObject())
+                                                                .projection(query.getFieldsObject())
+                                                                .update(((UpdateOpsImpl<T>) operations).getOps())
+                                                                .getOptions());
 
         return res == null ? null : mapper.fromDBObject(this, query.getEntityClass(), res, createCache());
 
--- a/morphia/src/test/java/org/mongodb/morphia/TestDatastore.java
+++ b/morphia/src/test/java/org/mongodb/morphia/TestDatastore.java
@@ -354,7 +354,7 @@ public class TestDatastore extends TestBase {
                                            .field("username").equal("john doe");
         UpdateOperations<FacebookUser> updateOperations = getDs().createUpdateOperations(FacebookUser.class)
             .inc("loginCount");
-        FacebookUser results = getDs().findAndModify(query, updateOperations);
+        FacebookUser results = getDs().findAndModify(query, updateOperations.isolated());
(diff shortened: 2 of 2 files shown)


morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
```diff
@@ -344,11 +344,16 @@ void process(final MappedClass mc, final Validation validation) {
         }
 
         updateForVersioning(query, operations);
-        DBObject res = dbColl.findAndModify(query.getQueryObject(), options.copy()
-                                                                           .sort(query.getSortObject())
-                                                                           .projection(query.getFieldsObject())
-                                                                           .update(((UpdateOpsImpl<T>) operations).getOps())
-                                           .getOptions());
+        DBObject queryObject = query.getQueryObject();
+        if (operations.isIsolated()) {
```
jyemin: My reading of $isolated is that it can safely be ignored for findAndModify, as that command only ever updates a single document.

jyemin: OK, close it as unnecessary.

evanchooly:
works as designed.

Issue #1054, opened 2016-10-26
$dec supports number argument


crzo:
Hey there,

is there any plan to build a convenient overloaded interface + implementation such as:

dec(String field, Number amount) ?

For now i am forced to use inc(String field, Number amount) where the amount must be negative. Its okay for now, but somehow confusing. Morphia can of course use the inc(String, Number) method under the hood, but on the business layer i don't want to "invert all values".

Cheers,
Chris

evanchooly:
Seems reasonable.  And simple enough.

UlrichCech:
Have provided a merge-request for this issue

evanchooly:
Thanks, @UlrichCech, for the PR.

PR #1057, opened 2016-10-26
#1054 $dec supports number argument

UlrichCech:
- added support for the dec-operation with a number as argument


UlrichCech:
accidental slip

Issue #1053, opened 2016-10-26
MongoClient's write concern is never respected


leoshaw:
Hi, not sure if I'm missing something, but it seems the underlying MongoClient's write concern is never respected, e.g.

```
  final Morphia morphia = new Morphia();
  final MongoClient client = new MongoClient("localhost",
      MongoClientOptions.builder().writeConcern(WriteConcern.MAJORITY).build());
  final Datastore datastore = morphia.createDatastore(client, "morphia_example");
  datastore.save(entity);
```

The save will happen with the datastore's default write concern, WriteConcern.ACKNOWLEDGED.  This wasn't the behaviour I was expecting - maybe if I hadn't set a preference in the MongoClientOptions then the datastore's default should apply, but I don't think that should be the case when a preference is specified in the options.

I realise it's simple to get the behaviour I want by setting the default write concern on the datastore, it's more about expected behaviour.  In my case it meant I had ACKNOWLEDGED for the write concern and MAJORITY for the read concern, which led to some interesting results!

This applies in 1.2.1

Regards,

Leo

PR #1061, opened 2016-10-31
Take into account MongoClient's write concern (#1053)

fiery-phoenix:


evanchooly:
Looks good.  thanks for the patch.

fiery-phoenix:
you are welcome

PR #1066, opened 2016-11-07
Deprecate Morphia#useBulkWriteOperations

### LINKED ISSUE #1052: Deprecate Morphia#useBulkWriteOperations
This field provides no real value since only bulk inserts are exposed and not mixed operations.  In such scenarios bulk write or not makes no difference.  If a future API exposes the ability to do mixed operations in bulk, this field would be redundant anyway.

evanchooly:
fixes #1052

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/Morphia.java
+++ b/morphia/src/main/java/org/mongodb/morphia/Morphia.java
@@ -19,6 +19,8 @@ import com.mongodb.DBObject;
 import com.mongodb.MongoClient;
 import org.mongodb.morphia.annotations.Embedded;
 import org.mongodb.morphia.annotations.Entity;
+import org.mongodb.morphia.logging.Logger;
+import org.mongodb.morphia.logging.MorphiaLoggerFactory;
 import org.mongodb.morphia.mapping.Mapper;
 import org.mongodb.morphia.mapping.MappingException;
 import org.mongodb.morphia.mapping.cache.EntityCache;
@@ -35,8 +37,8 @@ import java.util.Set;
  * @author Scott Hernandez
  */
 public class Morphia {
+    private static final Logger LOG = MorphiaLoggerFactory.get(Morphia.class);
     private final Mapper mapper;
-    private boolean useBulkWriteOperations = false;
 
     /**
      * Creates a Morphia instance with a default Mapper and an empty class set.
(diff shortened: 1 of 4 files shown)


morphia/src/main/java/org/mongodb/morphia/Morphia.java
```diff
@@ -160,19 +165,23 @@ public boolean isMapped(final Class entityClass) {
...
      * Configures Morphia to use bulk writes.  Only useful with MongoDB 2.6+.
      *
      * @param useBulkWriteOperations true if Morphia should use bulk writes
+     * @see <a href="https://github.com/mongodb/morphia/issues/1052">Issue #1052</a>
+     * @deprecated Setting this value has no value functionally or performance-wise.
      */
+    @Deprecated
     public void setUseBulkWriteOperations(final boolean useBulkWriteOperations) {
-        this.useBulkWriteOperations = useBulkWriteOperations;
+        LOG.warning("Setting useBulkWriteOperations has no value functionally or performance-wise.");
```
jyemin: I think the deprecate is sufficient.  Why also log a warning?

morphia/src/main/java/org/mongodb/morphia/Morphia.java
```diff
@@ -160,19 +165,23 @@ public boolean isMapped(final Class entityClass) {
     }
 
     /**
-     * @return true if Morphia should use bulk write operations.  Only useful with MongoDB 2.6+.
+     * @return false.  Setting this value has no value functionally or performance-wise.
+     * @see <a href="https://github.com/mongodb/morphia/issues/1052">Issue #1052</a>
      */
     public boolean isUseBulkWriteOperations() {
```
jyemin: This should also be deprecated then

morphia/src/main/java/org/mongodb/morphia/Morphia.java
evanchooly: Visibility mainly.  An ongoing warning that things have changed.  Figured it was the polite/correct thing to do.

morphia/src/main/java/org/mongodb/morphia/Morphia.java
evanchooly: Meant to and missed it.

morphia/src/main/java/org/mongodb/morphia/Morphia.java
jyemin: Please pull it out then, as it's inconsistent with all the other deprecated methods, none of which log a warning.

evanchooly: done

jyemin:
lgtm

PR #1068, opened 2016-11-10
Deprecation and warnings cleanup

### LINKED ISSUE #1067: Update tests to include new and deprecated versions
With all the new deprecations the tests could use a bit of clean up.  The task here is to duplicate every test exercising a deprecated method.  The new tests should be updated to use the new methods and the old tests should suppress the deprecated warnings.

evanchooly:
A lot of files changed here but it's mostly transitioning from the deprecated methods to the new and adding in any explicit tests needed to ensure coverage of the old methods.  Theses changes serve to eliminate build type warnings.

fixes #1067


morphia/src/main/java/org/mongodb/morphia/dao/BasicDAO.java
```diff
@@ -24,17 +25,19 @@
  * @author Olafur Gauti Gudmundsson
  * @author Scott Hernandez
  */
-@SuppressWarnings("deprecation")
+@SuppressWarnings({"WeakerAccess", "deprecation", "unused"})
```
rozza: Could you explain to me the new annotations and the causes?  I'm unfamiliar with "WeakerAccess" and haven't come across it before.

morphia/src/main/java/org/mongodb/morphia/dao/BasicDAO.java
```diff
@@ -57,7 +60,7 @@ public BasicDAO(final Class<T> entityClass, final MongoClient mongoClient, final
      * @param ds          the Datastore which gives access to the MongoDB instance for this DAO
      */
     public BasicDAO(final Class<T> entityClass, final Datastore ds) {
-        this.ds = (DatastoreImpl) ds;
+        this.ds = (org.mongodb.morphia.DatastoreImpl) ds;
```
rozza: Why the long reference? Are there two `DatastoreImpl` classes now?

morphia/src/main/java/org/mongodb/morphia/dao/BasicDAO.java
evanchooly: IDEA suggests that some methods could have weaker access, e.g., `protected` -> `private`.  Because this change would be unrelated and possibly breaking in some cases, this suppresses that suggestion.

morphia/src/main/java/org/mongodb/morphia/dao/BasicDAO.java
evanchooly: you can `@SuppressWarning` for *uses* of deprecated items but the compiler will still complain about imports.  Moving those uses to fully qualified class names circumvents that bug.

PR #1069, opened 2016-11-10
Complete the support for DeleteOptions

evanchooly:
introduces new overloads for deletes with options
deprecates methods that take a WriteConcern directly

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
+++ b/morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
@@ -121,6 +121,23 @@ public interface AdvancedDatastore extends Datastore {
      * kind}),
      * this method will delete any entity with the given {@code id}, regardless of its type.
      *
+     * @param kind    the collection name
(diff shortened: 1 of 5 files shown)


morphia/src/main/java/org/mongodb/morphia/AdvancedDatastore.java
```diff
@@ -121,14 +121,33 @@
      * kind}),
      * this method will delete any entity with the given {@code id}, regardless of its type.
      *
+     * @param kind    the collection name
+     * @param clazz   the Class of the entity to delete
+     * @param id      the value of the ID
+     * @param options the options to use when deleting
+     * @param <T>     the entity type
+     * @param <V>     is the type of the ID, for example ObjectId
+     * @return the result of this delete operation.
```
rozza: Do you have the convention to annotate with `@since` in Morphia? If so all these new methods need them!

evanchooly: yep!  missed those.  thanks.

PR #1072, opened 2016-11-15
fixes #1071

### LINKED ISSUE #1071: disableValidation doesn't work for fields using dot notation
When using disableValidation to turn off validation for a query, my expectation is that I should be able to pass whatever fields I want that would work against a basic mongo query and not have the system throw exceptions.  My specific use case is an embedded map that's not saved through morphia, but exists in the DB object.  With validation disabled, the QueryValidator throws an exception:

`org.mongodb.morphia.query.ValidationException: The field 'foo.bar' could not be found in 'CLASS'`

This is happening in 1.3.1-rc0, and it appears as if maybe it was caused by a null check inserted around line 94 of QueryValidator for issue #881.

I think that code should (instead) read:

```java
    if (mf == null && (mc.isInterface() || !validateNames)) {
        break;
    } else if (mf == null) {
        throw new ValidationException(format("The field '%s' could not be found in '%s'", prop, mc.getClazz().getName()));
    }
```

A simple test case is below.

```java
    @Test
    public void getSaveAndLoadNestedCollections() throws Exception {
        getMorphia().map(Class1.class);
        getDs().ensureIndexes(true);

        getDs().getDB().getCollection("user").save(
            new BasicDBObject()
                .append("@class", Class1.class.getName())
                .append("value1", "foo")
                .append("someMap", new BasicDBObject("someKey", "value"))
        );

        Query<Class1> query = getDs().createQuery(Class1.class);
        query.disableValidation().criteria("someMap.someKey").equal("value");
        Class1 retrievedValue = query.get();
        Assert.assertNotNull(retrievedValue);
        Assert.assertEquals("foo", retrievedValue.value1);
    }

    @Entity(value = "user", noClassnameStored = true)
    public static class Class1 {
        @Id
        private ObjectId id;

        private String value1;

    }
```

rlodge:
Would really like this in 1.3.0, if it's possible.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/query/QueryValidator.java
+++ b/morphia/src/main/java/org/mongodb/morphia/query/QueryValidator.java
@@ -91,7 +91,7 @@ final class QueryValidator {
                                                              + " validating - %s", part, mc.getClazz().getName(), prop));
                     }
 
-                    if (mf == null && mc.isInterface()) {
+                    if (mf == null && (mc.isInterface() || !validateNames)) {
                         break;
                     } else if (mf == null) {
                         throw new ValidationException(format("The field '%s' could not be found in '%s'", prop, mc.getClazz().getName()));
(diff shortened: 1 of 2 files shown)


evanchooly:
Looks good.  Thanks for the patch.

PR #1076, opened 2016-11-18
Iterator factory.  Resolves #1075

### LINKED ISSUE #1075: Create an IteratorFactory similar to QueryFactory
Similar to how Query is an interface and the QueryImpl class is basically only internal implementation and there is a factory on the Datastore for creating Query instances, it would be great to have MorphiaIterator and MorphiaKeyIterator to be interfaces with internal implementations with a factory interface on the Datastore to create them.  This would give users the option of adding functionality to the iterators without having to override and copy/paste a bunch of internal code.  (Our use case is gathering metrics from the iterators and sending them to graphite, but one could imagine other use cases).

rlodge:


evanchooly:
I'm unsure of this particular feature.  It adds complication for little gain.  The java driver already exposes an API to monitor its performance.  The mapping code is about to undergo a drastic update and, at best, this should wait until that is done.  I'm going to close this for now.

Issue #1006, opened 2016-09-16
add support for $sample


evanchooly:
[JAVA-1937](https://jira.mongodb.org/browse/JAVA-1937)

ankita-dhyani:
+1 any plans on releasing this feature ?

r3tina:
Hi, any update on when this will be included in a release? That would save me updating my own fork :-)

evanchooly:
I'm trying to clarify the plan with Mongo, myself...

evanchooly:
fixed by #1081

PR #1081, opened 2016-11-28
implemented $sample operation

r3tina:
small patch to support the $sample operation ( #1006 )


evanchooly:
This is great but should have some tests to confirm at least the shape of the pipeline document is correct.

r3tina:
I'll add a test in a few days.

r3tina:
So the "few days" became a few months. My apologies ,real life got in the way If you need anything else let me know.

Issue #1082, opened 2016-11-29
Create a configurable sanitizer for map keys


rlodge:
We operate Morphia in an environment where we frequently use custom string-like objects as map keys.  We need to be able to sanitize the keys of the maps to work with Mongo without having to override all the various map-to-mongo converters.

It would be great to have a configurable "map key sanitizer" that would by default do nothing, but could be overridden to ensure special characters aren't in the map keys.

rlodge:
I understand why sanitizing map keys in any specific way isn't a morphia concern exactly.  It's our concern, because we have a bunch of code that accepts URI's from some remote system and uses them as map keys. We're happy sanitizing them, but we don't want to hunt down a whole bunch of code over 75 different projects to do it at the point of origin. We could do it in a custom converter, but we only want the data changed for map keys, not for everywhere we store a URI.  So we need a way to change the strings when they're being used for map keys, but the various Converters don't provide a way for me to override that functionality specifically without creating new converters that duplicate or copy-paste a lot of the internal code.  It seemed fairly simple to introduce a sanitizer interface so that the behavior could be customized.  It makes sense to postpone this until after a major refactor, but it would be great if this kind of need (for extensibility of the core behavior of the library without having to re-implement or override private implementations) could be kept in mind for that refactor.

evanchooly:
I know this is years out of date but I'm going to close this one out.  Rather than introduce more infrastructure in to Morphia for a rather niche concern, this could be best implemented using `@EntityListeners` which would invoke whatever code you'd like to validate your map keys.

PR #1083, opened 2016-11-29
Implementation of issue #1082

rlodge:


evanchooly:
I'm not entirely convinvced this should be a Morphia concern though i could be talked in to it.  For now, though, I'm going to close this PR as the mapper code is about to see some signifcant refactoring.

Issue #1084, opened 2016-11-29
Make the constraint list on MappingValidator configurable


rlodge:
We have a need to slightly change the validation applied to map keys (in particular, the requirement that they be string-like).  In order to do this, I want to introduce a "ConstraintFactory" to "MappingValidator" so that MapperOptions can be used to configure the list of constraints that will be used for validation.

evanchooly:
I know this is old and probably pointlessly out of date, but what is the expectation here?  The server itself requires keys to be strings and Morphia just checks to make sure that's the case beforehand.  What *could* be done here is to check that the key is either a String or that a `Conversion` exists to/from that type to String such that keys could be properly persisted.  The `Conversion` API isn't exactly a public API but it's not an insurmountable problem.

rlodge:
We were using a fork of the project so we could get things like this, and we were looking for a way to get back to the main line.  We were using (for instance) a custom map serializable validator so that we could use custom annotations to handle maps that _didn't_ have string keys.  One of our problems with the library all along was that it wasn't customizable enough.  There was a lot of private API that we ended up forking to work around, and we were trying to get some more openness / configurability / overridability so we didn't have to use a fork; that was the basic justification for a couple of changes we requested.

That said, we are no longer using the library at all, so it's no longer important for _us_ to have this kind of flexibility.

evanchooly:
Fair enough.  Thanks for getting back to me.  I believe there's a test now that confirms this is done automatically provided the proper `Conversion`s are registered.  Sorry it took too long but the project is getting the attention it needs now so the backlog doesn't stagnate quite like it did then.

PR #1085, opened 2016-11-29
Implementation of issue #1084

rlodge:
Implements a constraint factory for the mapping validator so that users of the library can influence validation.  For #1084 .


evanchooly:
We're actively working on replacing the mapper with newer code with significantly different APIs.  Once that dust settles, we can readdress this idea.

Issue #1089, opened 2016-12-05
unsetting hashmap causes ArrayIndexOutOfBoundsException


umbreak:
Hello,

Updating to morphia 1.3.0 gives me an error when I try to unset a whole hashMap. The line it is:

```java
UpdateOperations<User> updateOpts = User.datastore.createUpdateOperations(User.class).disableValidation()
updateOpts.unset("services.$.crmConfig");
```
_Caused by: **java.lang.ArrayIndexOutOfBoundsException: 3**
	at java.util.Arrays$ArrayList.get(Arrays.java:3841)
	at org.mongodb.morphia.internal.PathTarget.next(PathTarget.java:96)
	at org.mongodb.morphia.internal.PathTarget.resolve(PathTarget.java:115)_

The model is as follows:

```java
    @Entity(value = "user", noClassnameStored = true)
    public class User {
        public List<Services> services;
        
        @Embedded
        public static class Services{
            public Map<CRMAttribute, Boolean> crmConfig;
        }
    }
```

Debugging, I can see that the problem comes in the method **resolve()** on the class **TargetPath**
You can see that, if you have the string `services.$.crmConfig`, the second time it runs into the loop, it throws ArrayIndexOutOfBoundsException. You need to add some extra checks for the method `hasNext()` in the else case:

```java
                if (!field.isMap()) {
                    translate(field.getNameToStore());
                } else {
                    next();  // consume the map key segment
                }
```
Cause I want to unset the whole map, so I'm not specifying any key, as assumed in that code

umbreak:
I'm not sure if it is enough just adding this check, or something else needs to be added.

fabriziobonavita:
Thanks for fixing it. This is blocking us at the moment, is there any release date for 1.3.1?

PR #1090, opened 2016-12-05
Update PathTarget.java

umbreak:
Adding an extra check to prevent ArrayIndexOutOfBoundsException in case the unset element it is a hashmap but the key is not specify. Related to issue: https://github.com/mongodb/morphia/issues/1089


evanchooly:
There should be tests to validate this.  See `PathTargetTest`.

umbreak:
See the issue linked. We definitely have errors updating (setting or unsetting) hashmaps, which we did not have in version 1.0.1.

evanchooly:
The question isn't whether or not you're seeing the error.  I trust that you are.  It's that your patch doesn't contain any tests to show that it fixes the error.  That's what I'd like to see before mergine.

evanchooly:
filed my own PR with the fix and tests

Issue #1091, opened 2016-12-06
IllegalArgumentException calling updateFirst


rlodge:
When calling the updateFirst method that takes a query and an entity (~line 865 in DatastoreImpl), I (always?) receive an IllegalArgumentException if the entity in question is versioned.  The exception I get is:

```
java.lang.IllegalArgumentException: Invalid BSON field name $inc
	at org.bson.AbstractBsonWriter.writeName(AbstractBsonWriter.java:516)
	at com.mongodb.DBObjectCodec.encode(DBObjectCodec.java:129)
	at com.mongodb.DBObjectCodec.encode(DBObjectCodec.java:61)
	...
	at com.mongodb.Mongo.execute(Mongo.java:845)
	at com.mongodb.Mongo$2.execute(Mongo.java:828)
	at com.mongodb.DBCollection.executeWriteOperation(DBCollection.java:342)
	at com.mongodb.DBCollection.update(DBCollection.java:567)
	at org.mongodb.morphia.DatastoreImpl.update(DatastoreImpl.java:1552)
	at org.mongodb.morphia.DatastoreImpl.update(DatastoreImpl.java:1507)
	at org.mongodb.morphia.DatastoreImpl.updateFirst(DatastoreImpl.java:869)
```

This appears to be because the `org.mongodb.morphia.DatastoreImpl#update(org.mongodb.morphia.query.Query<T>, com.mongodb.DBObject, org.mongodb.morphia.UpdateOptions, boolean)` method assumes that the input is an update operations document and not a replacement.

rlodge:
My concern about #1101 is that it changes the intent of the operation. By putting everything in a $set, it doesn't actually do a replace on the update.  Because the incoming DB object may not have all fields (since null ones might be absent, say), it will leave fields that aren't in the DBObject in place, only setting new ones.  I think that would make it (for instance) impossible to set a field to null using updateFirst.

evanchooly:
I did some testing prior to this.  It only ever updated the field listed in the update object.  Any existing fields in the document that weren't in the update document were unchanged.

rlodge:
[...]

PR #1094, opened 2016-12-06
Fix for #1091

rlodge:
Fix for #1091


evanchooly:
Thanks for the PR.  It was definitely in the right direction but didn't quite fix things correctly.  This approach requires a properly populated `version` field or you could end up in a scenario of reusing version numbers.  I've expanded on this PR a bit and opened a new one under #1101 if you want to take a look.

rlodge:
I"ll add a comment to #1091...

PR #1095, opened 2016-12-06
add a test to actually round trip an entity using BigDecimal

### LINKED ISSUE #1093: BigDecimal conversion problem - morphia 1.3 - MongoDB 3.4
Hello, i'm using MongoDB 3.4 and morphia 1.3 to test the new Decimal128 data type. I use to have a custom TypeConverter for my BigDecimals, they were stored on the database as strings.

So far i've disabled my custom converter, and updated the data directly on MongoDB. I can't make morphia save my BigDecimals as Decimal128. Inspecting the execution seems like the Mapper doesn't get the desired converter to BigDecimal, instead it uses the default Embedded Mapper.

The class for my test:

```java
public class Main {

    public static void main(String[] args) {
        Morphia morphia = new Morphia();
        Datastore datastore = morphia.createDatastore(new MongoClient("localhost"), "bigdecimal");
        Foo foo = new Foo();
        foo.setNumber(new BigDecimal("0.92348237942346239"));
        datastore.save(foo);
    }
}

@Entity
class Foo {

    @Id
    private ObjectId id;
    private BigDecimal number;

    public ObjectId getId() {
        return id;
    }

    public void setId(ObjectId id) {
        this.id = id;
    }

    public BigDecimal getNumber() {
        return number;
    }

    public void setNumber(BigDecimal number) {
        this.number = number;
    }

}
```

And a database `db.Foo.find()`:

```
{
        "_id" : ObjectId("5846fa9aa618770ee4f115d7"),
        "className" : "Foo",
        "number" : {
                "intVal" : {
                        "signum" : 1,
                        "mag" : [
                                21501499,
                                -1372597761
                        ],
                        "bitCount" : 0,
                        "bitLength" : 0,
                        "lowestSetBit" : 0,
                        "firstNonzeroIntNum" : 0
                },
                "scale" : 17
        }
}
```

Am I missing some configuration for the conversion BigDecimal <-> Decimal128 work properly?

evanchooly:
:facepalm:

fixes #1093

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
+++ b/morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
@@ -25,7 +25,7 @@ import java.math.BigInteger;
 /**
  * This provides a conversion to/from BigDecimal in applications using the Decimal128 type introduced in the 3.4.0 release of MongoDB.  This
  * converter also supports loading certain types of values from the database and converting them to BigDecimal.
- *
+ * <p>
  * Note:  While this is useful for gradual model migrations, the values will be saved as Decimal128 types when saved back.  This will
  * result in disparate types for a given field which may yield unexpected results when querying.
  *
@@ -34,6 +34,14 @@ import java.math.BigInteger;
  * @since 1.3
  */
 public class BigDecimalConverter extends TypeConverter implements SimpleValueConverter {
+
+    /**
+     * Creates the Converter.
+     */
+    public BigDecimalConverter() {
+        super(BigDecimal.class);
+    }
+
     @Override
     public Object decode(final Class<?> targetClass, final Object value, final MappedField optionalExtraInfo) {
         if (value == null) {
--- a/morphia/src/test/java/org/mongodb/morphia/converters/ConverterTest.java
+++ b/morphia/src/test/java/org/mongodb/morphia/converters/ConverterTest.java
@@ -17,10 +17,11 @@
 package org.mongodb.morphia.converters;
 
 import org.junit.Assert;
+import org.mongodb.morphia.TestBase;
 
 import static java.lang.String.format;
 
-class ConverterTest<F, T> {
+class ConverterTest<F, T> extends TestBase {
     private TypeConverter converter;
 
     ConverterTest(final TypeConverter converter) {
--- a/morphia/src/test/java/org/mongodb/morphia/converters/BigDecimalConverterTest.java
+++ b/morphia/src/test/java/org/mongodb/morphia/converters/BigDecimalConverterTest.java
@@ -17,8 +17,11 @@
 package org.mongodb.morphia.converters;
 
 import org.bson.types.Decimal128;
+import org.bson.types.ObjectId;
 import org.junit.Assert;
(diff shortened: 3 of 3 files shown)


morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
```diff
@@ -25,7 +25,7 @@
 /**
  * This provides a conversion to/from BigDecimal in applications using the Decimal128 type introduced in the 3.4.0 release of MongoDB.  This
  * converter also supports loading certain types of values from the database and converting them to BigDecimal.
- *
+ * <p>
```
rozza: Nit: unclosed tag - I think javadoc will complain

morphia/src/test/java/org/mongodb/morphia/converters/ConverterTest.java
```diff
@@ -17,10 +17,11 @@
 package org.mongodb.morphia.converters;
 
 import org.junit.Assert;
+import org.mongodb.morphia.TestBase;
 
 import static java.lang.String.format;
 
-class ConverterTest<F, T> {
+class ConverterTest<F, T> extends TestBase {
```
rozza: You probably should also add round trip test cases for your other ConverterTests to prevent future regressions.

morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
```diff
@@ -34,6 +34,14 @@
  * @since 1.3
  */
 public class BigDecimalConverter extends TypeConverter implements SimpleValueConverter {
+
+    /**
+     * Creates the Converter.
+     */
+    public BigDecimalConverter() {
```
jyemin: Let's unit test this constructor, and other similar converters.

evanchooly: #1098

morphia/src/test/java/org/mongodb/morphia/converters/ConverterTest.java
evanchooly: #1097

morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
evanchooly: nope.  no complaints.

morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
evanchooly: I'll update the commit message on the merge commit...

morphia/src/main/java/org/mongodb/morphia/converters/BigDecimalConverter.java
rozza: Ah I remember its Java 8 doclint that will complain, so it depends on which version you are using to build the javadoc and / or you have disabled doclint.

evanchooly:
Good to merge?

jyemin:
LGTM

PR #1102, opened 2016-12-07
when collecting index definitions, look at the parent classes as well…

### LINKED ISSUE #1086: Generates collections for abstract base classes of entities
Since version 1.3.0 Morphia also generates collections for abstract baseclasses of entities.

Example base class:

```java
@Converters(JodaDateTimeConverter.class)
@Indexes({
    @Index(fields = {@Field(value = AbstractValueSliceStorage.TIMESTAMP_FIELD), @Field(value = AbstractValueSliceStorage.SERIE_FIELD)}, options = @IndexOptions(unique = true))
})
public abstract class AbstractValueSliceStorage implements Serializable {
    private static final long serialVersionUID = 1L;
   ** code surpressed **
}
```

Example subclass:
```java
@Entity(value = "runtime.value.converted", noClassnameStored = true)
public class ConvertedMinutesInHour extends AbstractValueSliceStorage {
    private static final long serialVersionUID = 1L;

}
```

In 1.2.1 and before only the runtime.value.converted collection is created, in 1.3.0 morphia also creates a collection named AbstractValueSliceStorage.

evanchooly:
… as the nested types for definitions.

don't create indexes when the mapped class is abstract.

fixes #1086

### DIFF
--- a/morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
+++ b/morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
@@ -17,7 +17,6 @@
 package org.mongodb.morphia;
 
 import com.mongodb.BasicDBObject;
-import com.mongodb.DBCollection;
 import com.mongodb.DBObject;
 import com.mongodb.client.MongoCollection;
 import org.bson.BsonDocument;
@@ -37,7 +36,6 @@ import org.mongodb.morphia.annotations.Index;
 import org.mongodb.morphia.annotations.IndexOptions;
 import org.mongodb.morphia.annotations.Indexed;
 import org.mongodb.morphia.annotations.Indexes;
-import org.mongodb.morphia.annotations.Property;
 import org.mongodb.morphia.annotations.Text;
 import org.mongodb.morphia.mapping.MappedClass;
 import org.mongodb.morphia.mapping.Mapper;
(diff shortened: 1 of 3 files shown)


morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
```diff
@@ -109,13 +111,13 @@ public void calculateKeys() {
...
-        MappedClass mappedClass = getMorphia().getMapper().getMappedClass(IndexedClass.class);
+        String collectionName = getDs().getCollection(IndexedClass.class).getName();
+        MongoCollection<Document> collection = getDatabase().getCollection(collectionName);
+        Mapper mapper = getMorphia().getMapper();
 
-        indexHelper.createIndex(collection, mappedClass, false);
+        indexHelper.createIndex(collection, mapper.getMappedClass(IndexedClass.class), false);
         List<DBObject> indexInfo = getDs().getCollection(IndexedClass.class)
                                           .getIndexInfo();
-        assertEquals("Should have 5 indexes", 5, indexInfo.size());
```
rozza: How come this check was removed - shouldn't there be a fixed number of indexes?

evanchooly: Well my initial thought was we'd catch any extra indexes in the loop but that won't catch missing indexes so I'll restore that check.

morphia/src/test/java/org/mongodb/morphia/IndexHelperTest.java
```diff
@@ -129,22 +131,27 @@ public void createIndex() {
                 assertEquals(parse("{ 'nest.name' : 1} "), dbObject.get("key"));
             } else if (name.equals("searchme")) {
                 assertEquals(parse("{ 'text' : 10 }"), dbObject.get("weights"));
+            } else if (name.equals("indexName_1")) {
```
rozza: Checkstyle probably is complaining about the `"`

PR #1110, opened 2016-12-14
add checks before calling next() when evaluating path segments

### LINKED ISSUE #1089: unsetting hashmap causes ArrayIndexOutOfBoundsException

evanchooly:
fixes #1089

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/internal/PathTarget.java
+++ b/morphia/src/main/java/org/mongodb/morphia/internal/PathTarget.java
@@ -34,6 +34,7 @@ import static org.mongodb.morphia.internal.MorphiaUtils.join;
  * @since 1.3
  */
 public class PathTarget {
+    private final String path;
     private final List<String> segments;
     private boolean validateNames = true;
     private int position;
@@ -54,6 +55,7 @@ public class PathTarget {
         this.root = root;
         segments = asList(path.split("\\."));
         this.mapper = mapper;
+        this.path = path;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


evanchooly:
Correct.  It's a live exercise of the fix.  This operation ends up creating the fault path being checked in the update to `PathTargetTest`.

PR #1111, opened 2016-12-16
deprecate the `updateFirst` and `update*` methods

evanchooly:
This commit mostly just makes the `update` methods consistent with the changes to the saves and deletes.  The biggest arguable change is the deprecation of the `updateFirst` variants.  Looking at the API it seems there was some attempt at symmetry, over the years, to make `updateFirst` versions of the `update` methods.  This hasn't quite happened, however, and with easy access to `UpdateOptions` now, having `updateFirst` as separate methods makes less sense.  In the spirit of simplifying and streamlining the API, I've elected to deprecate the `updateFirst` methods in favor of the `update` methods.

The tests have been updated to use the correct methods and some new tests introduced to make sure the old methods still get tested.  I did a little bit of clean up to help isolate where we've been over testing and reduced that redundancy.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/Datastore.java
+++ b/morphia/src/main/java/org/mongodb/morphia/Datastore.java
@@ -725,7 +725,9 @@ public interface Datastore {
      * @param createIfMissing if true, a document will be created if none can be found that match the query
      * @param <T>             the type of the entity
      * @return the results of the updates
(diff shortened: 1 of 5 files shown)


evanchooly:
I'll merge this if there are no complaints then...

jyemin:
@rozza can you review please?

Issue #1123, opened 2017-01-11
UpdateOpsImpl does not respect fieldName annotations for maps


VPriesnitz:
Hello, 
This is a critical bug.

When updating a field of type Map that has a fieldName annotation (@Embedded or @Property), then an update on that field does not replace the field's name with the provided value. 

POJO
class Payload {
        @Id
        private ObjectId id;

        @Property("c")
        private final Map<String, String> content = new HashMap<String, String>();
    }

Update
`ds.createUpdateOperations(Payload.class)
                .set("content.key", "newValue");
`

results in mongo: 
`{
      "_id": { "$oid": "587609d12e934813a926ec69" }, 
      "c": { "key": "value" }, 
      "content": { "key": "newValue" }
}`
Note that the Document contains now both "c" and "content". The update returns without error and subsequent retrievals will ignore this update.   

This behaviour was introduced in commit  d0d9e04975717a1b74204a3b15183bfbd1d99f5d

Attached is patch that contains a unitTest to reproduce the issue.  (had to rename from .patch to .TXT got github )
[testUpdatingMapFieldsAnnotatedWithFieldNames.TXT](https://github.com/mongodb/morphia/files/698553/testUpdatingMapFieldsAnnotatedWithFieldNames.TXT)

PR #1124, opened 2017-01-11
always translate according to annotation + unit test to verify

danielBreitlauch:
This solves https://github.com/mongodb/morphia/issues/1123.


danielBreitlauch:
Why didn't you just accept the pull request instead of copying the code and committing it yourself?

evanchooly:
i moved the test to PathTargetTest and reused one of the test entities.

PR #1130, opened 2017-01-17
bugfix: the resolve function would be called multiple times

danielBreitlauch:
Small fix such that resolved is set to true.


evanchooly:
I think I see what's going on here but a bug description and tests would be really helpful.

danielBreitlauch:
It is just a performance regression. By adding the break; there 'resolved = true;' is never called. So the routine is not cached and executed every time.

danielBreitlauch:
@evanchooly do you consider this fix of the previously introduces performance regression?

danielBreitlauch:
Ping?

PR #1147, opened 2017-02-03
Added timezone to test so that it no longer fails when run in CET.

trishagee:
This test always fails for me locally, it shows an actual date of yesterday rather than today.  Adding a specific timezone appears to have fixed it for me.

### DIFF
--- a/morphia/src/test/java/org/mongodb/morphia/aggregation/AggregationTest.java
+++ b/morphia/src/test/java/org/mongodb/morphia/aggregation/AggregationTest.java
@@ -148,7 +148,7 @@ public class AggregationTest extends TestBase {
     @Test
     public void testDateToString() throws ParseException {
         checkMinServerVersion(3.0);
-        Date joined = new SimpleDateFormat("yyyy-MM-dd").parse("2016-05-01");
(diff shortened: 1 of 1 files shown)


udoheld:
The current master seems to be broken, the same test failed for me as well.

20:41:45.857 [DEBUG] [TestEventLogger] org.mongodb.morphia.aggregation.AggregationTest > testDateToString FAILED
20:41:45.858 [DEBUG] [TestEventLogger]     org.junit.ComparisonFailure: expected:<2016-0[5-01]> but was:<2016-0[4-30]>
20:41:45.858 [DEBUG] [TestEventLogger]         at org.junit.Assert.assertEquals(Assert.java:115)
20:41:45.858 [DEBUG] [TestEventLogger]         at org.junit.Assert.assertEquals(Assert.java:144)
20:41:45.858 [DEBUG] [TestEventLogger]         at org.mongodb.morphia.aggregation.AggregationTest.testDateToString(AggregationTest.java:165)


After pulling the fix it works in AEST as well.

trishagee:
@jyemin Pretty please can you check this and accept it?  The build will not work in a number of timezones (including the UK and Europe!) without this fix.

PR #1150, opened 2017-02-16
Projection expressions with single arguments e.g. {"$size": "$source"}

kothar:
After #667 was merged, expressions could be composed such as

`Projection.expression("$size", Projection.projection("field"))`

Which would be translated to:

`{"$size": ["$field"]}`

Though `$size` only accepts a single argument, an array containing a single string is also accepted.

After @e95cdc3271408000bc57cf8f94c9468de55f1c43 ([Location of change](https://github.com/mongodb/morphia/commit/e95cdc3271408000bc57cf8f94c9468de55f1c43#diff-b8c970e87e213b797b5f99cabb700305R262)) a check was added to unwrap single-argument expressions, but it assumes that the type of the argument is a `DBObject`. This fails in the above case, where the argument is a projection reference string.

This PR adds a helper method for `$size` projection expressions, and a corresponding test to make sure the argument is unwrapped correctly from the arguments list.

The fix for the `ClassCastException` moves the unwrapping check to the caller of `toExpressionArgs` so that the string value can be handled without a cast to DBObject causing an exception.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/aggregation/AggregationPipelineImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/aggregation/AggregationPipelineImpl.java
@@ -223,7 +223,6 @@ public class AggregationPipelineImpl implements AggregationPipeline {
      * @param projection the project to apply
      * @return the DBObject
      */
-    @SuppressWarnings("unchecked")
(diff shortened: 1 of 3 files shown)


evanchooly:
rebased and merged.
Thank you for the patch and I apologize for the delay.

PR #1163, opened 2017-05-01
remove morphiaiterator & morphiakeyiterator from public exp…

amanchoudhary:
…osed api and replacing it with iterator


evanchooly:
First, thanks for all the work.  I agree with the spirit of this change and have done similar work in the 2.0 branch.  But these are breaking changes in the 1.x line and so can not be merged.  This will happen in 2.0, though.

PR #1186, opened 2017-08-22
fix the call getCursor error

shrekxu:
When call the getCursor function, it will throw the following exception, just need to use the com.mongodb.Cursor to solve this problem.
java.lang.ClassCastException: com.mongodb.MongoCursorAdapter cannot be cast to com.mongodb.DBCursor
	at org.mongodb.morphia.query.MorphiaIterator.getCursor(MorphiaIterator.java:76)

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/query/MorphiaIterator.java
+++ b/morphia/src/main/java/org/mongodb/morphia/query/MorphiaIterator.java
@@ -1,6 +1,7 @@
 package org.mongodb.morphia.query;
 
 
+import com.mongodb.Cursor;
 import com.mongodb.DBCursor;
 import com.mongodb.DBObject;
 import org.mongodb.morphia.Datastore;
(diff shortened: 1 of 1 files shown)


shrekxu:
After this bug fixed, we can use the getCursor return the cursor to get the DBObject. it will very easy to get the Aggregation result without creating a defined Class.

jyemin:
While I agree that it's useful, this change breaks binary compatibility for any existing users of the getCursor method.  To avoid this problem, consider instead adding a new method, e.g.

    public Cursor getRawCursor() {
        return (Cursor) wrapped;
    }

I don't love the name, but the good name is already taken.  

What do you think @evanchooly ?

evanchooly:
i'm not entirely sure the "why" of this PR at this point.  In fact, that method's not even used by Morphia so I'm questioning why it's even there.  It *is* gone in the 2.0 code as this whole class goes away.  I'm going to close this PR for now but if this issue is still a bother if you could file an issue with a test case to reproduce we can take another look at it.

PR #1191, opened 2017-09-19
Fix wrong position of IndexOptions in doc

polo2ro:
options = @IndexOptions must be set in @Index and not in @Field


docs/reference/content/guides/indexing.md
```diff
@@ -25,7 +25,10 @@ has two primary components to cover here:  `fields` and `options`.  An index def
 @Entity
 @Indexes({
     @Index(fields = @Field("field2", type = DESC)),
```
jyemin: This is actually incorrect as well.  It should be:

         @Index(fields = @Field(value = "field2", type = DESC)),

Would you mind fixing that as well?

polo2ro: I added this to the branch

PR #1222, opened 2018-01-19
performance improvement: memoized some logically static properties

### LINKED ISSUE #1221: Query performance could be better if expensive mappings are memoized
This is a request for performance improvements. 
The issue is noticeable especially for bigger java object, in terms of total number of mapped fields. 

I suggest to memoize 
- java FieldsNames -> mongo keys
- the typeConverters for classes and MappedFields 
- the CustomMapper for a MappedFields

Some experiments show a potential performance increase of around 50% (this depends of course heavily on the actual usecase) 

Regards,
Vincenz

VPriesnitz:
this resolves #1221 

I added an adhoc unitTest that displays the performance differences. 
It is "@Ignored" as to not slow down the build.


evanchooly:
I like the numbers but org.mongodb.morphia.ext.ExternalMapperExtTest#testExternalMapping is currently failing with this patch.  Can you dig in to why?

evanchooly:
looks like you need a rebase...

VPriesnitz:
i think i have addressed all issues, could you please take another look?

PR #1227, opened 2018-01-30
Simple interface for discriminator persistence

jwilner:
Hi! I'd like to propose adding two interfaces for a rudimentary polymorphic discriminator system to Morphia. 

### background

Morphia currently handles polymorphic deserialization by storing a fully qualified class name in a hardcoded `className` field. Although simple and predictable, this method imposes serious restrictions on clients of the objects. Repackaging a class requires jumping through hoops in production, and non-Java clients of the serialized data types are at a disadvantage. 

### similar solutions to the problem

Other Java serialization libraries like Jackson and JAXB provide for logical names as discriminator values, which provide greater flexibility via their indirection. See, for example, Jackson's `@JsonTypeInfo` [annotation ](http://fasterxml.github.io/jackson-annotations/javadoc/2.9/) with its `Use.NAME` enum name. 

### proposed code

In this pull request, I encapsulated the `className` field and its class loader logic with two interfaces -- a `ClassInfoPersister` and a `ClassIdMapper`. The `ClassInfoPersister` is tasked with, well, persisting class information; it provides methods for storing class information, removing class information, adding class information to a projection, and lastly, loading a `Class<T>` from data. The `ClassIdMapper` is a simpler interface to implement; it just goes from `String` class id to `Class<T>` and vice versa. 

### expected usage

Many type discriminator strategies will still use one fixed field -- `className` or `@type` -- for their class id; in this case, they need only implement `ClassIdMapper` and parameterize `DefaultClassInfoPersister`. On the other hand, if you want to implement a more complicated discriminator pattern -- e.g. dispatching off a preexisting enum field -- you'll likely need to implement the larger interface. I also provided a pretty common sense implementation of `ClassIdMapper` that is backed by a simple map (I'm going to use this one immediately in some other code).

With an eye towards more complicated future implementations, most of the `ClassInfoPersister` methods include a `Class<?>` for any known supertypes currently available (e.g. either the type of a `MappedField` or an entity), although the provided implementation ignores those values.

### backwards compatibility

The default implementation is completely backwards compatible. There are no changes to `public` or `protected` APIs; I did however take the liberty of deprecating three members:
- the public static constant `Mapper#CLASS_NAME_FIELDNAME`, which is now more properly owned by `DefaultClassInfoPersister`. 
- `public Map<String, Class> getClassNameCache` -- this seemed intended to expose state for testing but no longer directly owns the state with the above intervening interfaces
- a protected helper method for accessing a class loader that's no longer used internally; I doubt there are any clients out there, but who knows?

### from here

An obvious next step in Morphia's style would be to add an annotation that indicates the discriminator value. In its mapping process, Morphia could detect the alternate `Discriminator` anno, add the value and the annotated class to a hash map, and then pass it to the above configured `FixedMapClassIdMapper` -- letting clients annotate define their own logical names via annotations, just like other libraries.

Let me know what you think! I'd find this pretty useful.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
+++ b/morphia/src/main/java/org/mongodb/morphia/DatastoreImpl.java
@@ -442,7 +442,7 @@ public class DatastoreImpl implements AdvancedDatastore {
 
         Object id = key.getId();
         if (id instanceof DBObject) {
-            ((DBObject) id).removeField(Mapper.CLASS_NAME_FIELDNAME);
+            mapper.getOptions().getClassInfoPersister().removeClassInfo((DBObject) id);
         }
         return get(clazz, id);
     }
--- a/morphia/src/main/java/org/mongodb/morphia/ObjectFactory.java
+++ b/morphia/src/main/java/org/mongodb/morphia/ObjectFactory.java
@@ -1,6 +1,7 @@
 package org.mongodb.morphia;
 
 import com.mongodb.DBObject;
+import org.mongodb.morphia.mapping.classinfo.ClassInfoPersister;
 import org.mongodb.morphia.mapping.MappedField;
 import org.mongodb.morphia.mapping.Mapper;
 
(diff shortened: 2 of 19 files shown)


evanchooly:
First, thank you for all the thought and work you've put in to this.  And secondly, I apologize for the lag in a response.  I have completely redone how mapping and polymorphism is done in what will be the 2.0 branch in a matter of days, though, and so these changes, apart from the basic git conflicts, would no longer apply.  I'm going to close out this PR but I encourage you to take a look at the `mapper` branch.  This is where my refactor currently lives and I will be migrating changes from there to a new 2.0.0 branch in the coming days.

Issue #1008, opened 2016-09-16
add support for $lookup stage 


evanchooly:
[JAVA-1936](https://jira.mongodb.org/browse/JAVA-1936)

PR #1228, opened 2018-02-08
#1008 Support for sub queries in $lookup

telomir:
This pull request aims to provide support for the new 3.6 sub-queries in the $lookup aggregation stage as documented in https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/#join-conditions-and-uncorrelated-sub-queries. Test case is taken from the official documentation at: https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/#join-conditions-and-uncorrelated-sub-queries

I was and still am really unsure about the interface because of the let parameter being just a BasicDBObject. It's usage came in quite handy for me, at least handier than a Map<String, String>. As I am not that deep into the depths of Morphia interfaces, I might have overlooked any similar interface where one would have to use a simple map. Happy to refactor that accordingly.

### DIFF
--- a/morphia/src/main/java/org/mongodb/morphia/aggregation/AggregationPipeline.java
+++ b/morphia/src/main/java/org/mongodb/morphia/aggregation/AggregationPipeline.java
@@ -130,8 +130,20 @@ public interface AggregationPipeline {
      */
     AggregationPipeline lookup(String from, String localField, String foreignField, String as);
 
-    // TODO: I am unsure about the interface, especially with the let parameter type.
-    // TODO Documentation
+    /**
+     * Performs a left outer join to an unsharded collection in the same database to filter in documents from the “joined” collection for
+     * processing. To each input document, the $lookup stage adds a new array field whose elements are the matching documents from the
+     * “joined” collection. The $lookup stage passes these reshaped documents to the next stage.
+     *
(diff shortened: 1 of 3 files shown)


telomir:
This feature could really profit from an implementation of [$expr](https://docs.mongodb.com/manual/reference/operator/query/expr/#op._S_expr), but I could not yet wrap my head around it how to implement that with an easy-to-use interface and without writings tons of code today.

telomir:
The test case should be refactored after #1229 is resolved.

evanchooly:
I've almost merged this twice before remembering about all the `BasicDBObject` bits in the API.  It's not a clean API and the 2.x branch all but completely removes `DBObject` from every aspect of the API so I'm loathe to add a new instance here.  The expression of the subqueries exposes too much of the inner workings for my taste, too, so I'd like to give that some thought as well.  My gut tells me the current API, irrespective of this patch, doesn't provide the appropriate hooks to write a more user friendly API just yet.

evanchooly:
This PR won't work with the new agg bits i'm wrapping up but this will help remind me to add these bits.

evanchooly:
This has been done in the new API.  you can now specify a pipeline with the lookup stage.

PR #1279, opened 2018-10-31
Retain order of Map in update flow

shrey-garg:
When calling DAO<T, U>.update(query, operation) and setting a LinkedHashMap in an attribute would result in ordering of that Map being lost due to the usage of HashMap.

I have replaced this usage with a LinkedHashMap and this fixes this issue.


evanchooly:
Good catch.  Thanks for the update.

PR #1281, opened 2018-11-01
Retain order of Map in update flow [1.4.x]

shrey-garg:


evanchooly:
actually there are no other 1.4.x releases planned.  1.5.0 will be the next release.

PR #1394, opened 2019-10-28
Add classLoader to mapper options builder

jonahseguin:
This is necessary for some use cases (for example, when working with Spigot/Bukkit JavaPlugins) (see https://spigotmc.org) when the `PluginClassLoader` is used.  In order to prevent `NoClassDefFoundError` for entity classes, the `JavaPlugin`'s ClassLoader needs to be used, and I'm sure this is applicable to other applications as well.

**Example of my use in the previous version (when it was possible to override the DefaultCreator's `getClassLoaderForClass()` method):**

```java
MapperOptions options = morphia.getMapper().getOptions();
        morphia.getMapper().setOptions(MapperOptions.builder(options)
                .objectFactory(new DefaultCreator(options) {
                    @Override
                    protected ClassLoader getClassLoaderForClass() {
                        return myClassLoader;
                    }
                })
                .build());
```

**Example usage with this commit:**

```java
morphia.getMapper().setOptions(MapperOptions.builder(morphia.getMapper().getOptions())
                .classLoader(myClassLoader)
                .build());
```

### DIFF
--- a/morphia/src/main/java/dev/morphia/mapping/MapperOptions.java
+++ b/morphia/src/main/java/dev/morphia/mapping/MapperOptions.java
@@ -66,6 +66,7 @@ public class MapperOptions {
         mapSubPackages = options.isMapSubPackages();
         dateStorage = options.dateStorage;
         disableEmbeddedIndexes = options.disableEmbeddedIndexes;
+        classLoader = options.getClassLoader();
     }
 
     private MapperOptions(final Builder builder) {
@@ -83,6 +84,7 @@ public class MapperOptions {
         referenceMapper = builder.referenceMapper;
         valueMapper = builder.valueMapper;
         dateStorage = builder.dateStorage;
+        classLoader = builder.classLoader;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


evanchooly:
Have you tried this with 1.5.7?  It was just published in the last day or two and I've done a fair bit of work in this exact area.

jonahseguin:
@evanchooly Yeah, just tried with 1.5.7 -- the classLoader variable is read-only in the MapperOptions class and there is no option for it in the builder

evanchooly:
and it's *still* necessary?  That's disappointing...

Not the end of the world, though, i guess.  It's a nice escape hatch, I suppose, for when morphia can't quite figure it all out.  I'll take a look in the next day or two then.

jonahseguin:
@evanchooly Yeah, just for some use cases where we need to provide our own class loader.
Thanks

evanchooly:
I just pushed 1.5.8 so you should see this on central in the next hour or two.

PR #1404, opened 2019-11-28
Update "latest documentation" link to 1.5

tamasperger:
Update "latest documentation" link on the landing page to 1.5


evanchooly:
Hrm.  All those should have been auto updated during the release process.  I'll look in to it.  Thanks for catching and patching that.

PR #1531, opened 2020-11-13
Bump mongo driver from 4.0.3 up to 4.1.1

a-panchenko:


evanchooly:
I can apply this patch but i'm hoping to not need any more releases in 1.6.x.  My hope is that folks migrate to 2.0.  Is this impeding you in any way?

evanchooly:
had a bug come up elsewhere so there'll be a 1.6.1 after all.

PR #1535, opened 2020-11-28
new config format, and use sonatype API to fetch SNAPSHOT javadoc jar.

djencks:
This shows the new @djencks/antora-javadoc config format and also how to use the sonatype REST api to fetch the latest SNAPSHOT javadoc jar. If you want to keep the configuration in the descriptors rather than the playbook, I think it would be easier to just modify the config yourself than deal with a PR per branch.  If you'd rather have PRs let me know.

### DIFF
--- a/docs/antora.yml
+++ b/docs/antora.yml
@@ -19,5 +19,5 @@ extensions-config:
   - path: '@djencks/antora-javadoc'
     config:
       sources:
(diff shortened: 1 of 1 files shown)


evanchooly:
Thanks!  I'll push the change to the other branches.

PR #1563, opened 2021-01-28
Bug/deserialization fails java sql timestamp

ptwohig:
It looks like the driver deals only in ```java.util.Date``` instances. In my production code, I use ```java.sql.Timestamp```. It seems that the driver really has no way of knowing whether the client code wants the value to be fetched later as a ```java.sql.Timestamp``` (or any other subclass of ```java.util.Date``` for that matter).

What this change does is use a special ```DateFieldAccessor``` and constructs the desired type by reflectively invoking a constructor accepting long. (ie ```DateSubclass(long)```) which should handle all of the ```java.util.Date``` subclasses in the JRE.


ptwohig:
This is a goofy fix. Gonna cancel this out.

PR #1564, opened 2021-01-29
Bugfix/timestamp codec

ptwohig:
I think this is a better fix than what I had previously made. This is basically what I put in my production code to work around the current issue.


evanchooly:
discussed in irc.

Issue #1586, opened 2021-03-19
Wrong deserialization of polymorphic entities with lifecycle 


renat-nosto:
**Describe the bug**
If there is 
```java
    @Entity(value = "entity")
    static class BaseEntity{
        @Id
        ObjectId id;

        @PostLoad
        void audit() {
        }
    }

    static class ChildEntity extends BaseEntity {
        String value;
    }
```

```java
ChildEntity childEntity = new ChildEntity();
ds.save(childEntity);
BaseEntity saved = ds.find(BaseEntity.class).filter(Filters.eq("_id", childEntity.id)).first();
```
yields `BaseEntity` in presence of lifecycle, correct behaviour yielding `ChildEntity` is observed in absence of  `@PostLoad`

** Please complete the following information: **
 - Server Version: 4.2.8
 - Driver Version: 4.0.5
 - Morphia Version: 2.1.4

evanchooly:
First, thanks the for the issue and the PR.  *But*, you filed the bug against 2.1.4 but the PR against master which is significantly different than how 2.1.x works.  so I think what i'll do here is bump this issue to 2.2.0 and create my standard "backport issue #blah" for 2.1.x.  We can't just cherry pick the PR because that code is very different.

PR #1587, opened 2021-03-19
#1586 deserialization of polymorphic entities with lifecycle

renat-nosto:
respect discriminator when deserializing entity with lifecycle


renat-nosto:
Hello, @evanchooly this is a fix for #1586, I've fixed the style issue, updated the code to have better alignment between lifecycle and non-lifecycle case (do not ignore wrong discriminator) and expanded the test a bit and feel like the PR now is in a good shape to be merged.

PR #1598, opened 2021-04-15
Fixed variable typo in transactions.adoc

brayce1996:
Fixed variable typo.
Based on the context, the variable `account` might be `user`.


evanchooly:
good catch.  thanks for the fix.

PR #1646, opened 2021-09-29
Use alternative instantiation

richarddd:
This solves Kotlin data classes and avoids using unsafe for most of the cases. Falls back to unsafe allocator as a last resort.

Fixes: https://github.com/MorphiaOrg/morphia/issues/1645


evanchooly:
Thanks for the PR.  Not everyone puts in the effort (or can!) so I really do appreciate it.  I have a fix for this in place for [2.3.0](https://github.com/MorphiaOrg/morphia/issues/1640) if you want to try a snapshot build to confirm it works in your case.  In the meantime, i'm going to let this PR sit for a bit until we see how things shake out.  I'll probably end up doing a back port of the 2.3.0 changes in this case.

evanchooly:
this is solved in 2.3.0

Issue #1648, opened 2021-10-01
Significant performance decline with @Reference(idOnly = true) migrating from 1.4.1


ugljesas:
I have a project that uses Morphia 1.4.1 that I'm looking to migrate to 2.2.

Running performance tests, I've found that certain workloads have degraded tremendously compared to the previous version. 

Doing some debugging I figured out it's connected to fetching multiple objects through query.iterator().toList() (previously query.asList()) that contain Reference(idOnly = true).

In the case of Lazy references, I've done a controlled test and only fetched the _id and the reference field, without actually accessing it, and there seems to be significant overhead compared to the previous version and fetching it as a plain ObjectId.

Non-lazy references seem to add even more overhead and performance degrades further compared to what I have in production. 

The consequence is that a request that fetches about 500 objects takes about 300ms with 1.4.1, but blows up to 2.5s with 2.2.1 on the same env. As a quick test, I've run the same workload using plain ObjectIds for references and handling them internally and the performance is about the same.

I would love some input on why this might be happening and if there is a workaround or fix in the works?

Server Version: 3.6.19
Driver Version: 4.2.2
Morphia Version: 2.2.1

evanchooly:
I won't hold the 2.2.2 release for this one but I've been intending to set up a benchmarks project for a while now to track such regressions.  I'll make this the inaugural benchmark.

h-arlt:
We observed performance degradation as well after migrating to Morphia v2. 
@evanchooly Did you find some time for further investigation and/or do have any updates on this?

h-arlt:
Once we deployed Morphia v2 into our production system, CPU usage doubled in comparison to v1. After doing some profiling, we noticed that there were a lot of calls to `ReferenceCodec#createProxy()` and some of them took 6s or more. So we took a closer look at this method and realized that Proxy classes are generated for each and every instance of a given Reference type.

Since byte code generation and class loading is rather expensive, this should only be done once per Reference type and not once per instance. This [PR](https://github.com/MorphiaOrg/morphia/pull/1810)  adds caching of generated proxy classes to `ReferenceCodec` which improved overall performance in our production system significantly.

PR #1810, opened 2022-01-27
#1648: Cache generated proxy classes

h-arlt:
This PR adds caching of generated proxy classes to `ReferenceCodec` in order avoid superfluous byte code generation and class loading which is quite expensive when done for each and every instance of a reference type.

Having caching in place improved overall performance in our production system significantly and should help other users fighting with poor performance when using `@Reference(idOnly=true)` as well.

Changes to `ReferenceCodec`:
- Generate proxy class for given reference type only once instead for each instance of that reference type
- Proxy classes are cached by Byte Buddy's type cache and store the respective invocation handler as instance field value

### DIFF
--- a/core/src/main/java/dev/morphia/mapping/codec/references/ReferenceCodec.java
+++ b/core/src/main/java/dev/morphia/mapping/codec/references/ReferenceCodec.java
@@ -26,10 +26,14 @@ import dev.morphia.mapping.lazy.proxy.ReferenceException;
 import dev.morphia.query.QueryException;
 import dev.morphia.sofia.Sofia;
 import net.bytebuddy.ByteBuddy;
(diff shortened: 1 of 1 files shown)


evanchooly:
Thank you for this.  I was literally just looking at this code the other day and had the same realization.  I haven't had time to build out that cache but it looks like you did.  :)  I'll take a look at the PR and see about it getting it merged hopefully this weekend.

h-arlt:
You're welcome! I stumbled upon a similar [question/issue](https://github.com/raphw/byte-buddy/issues/840) in the Byte Buddy project and the Byte Buddy owner himself proposed that solution that I just had to implement ;)
Thanks for your feedback and the efforts you put into this framework!!

PR #2171, opened 2022-12-20
Enum collection and map codecs

daniel-exceed:
enum map and enum set are very efficient in storage and performance in java. they don't have a regular constructor, so a special handling is needed to instantiate them


evanchooly:
These types should be handled by their own codecs.  That codec is intended to help resolve types involving the Collections and Map interfaces.

daniel-exceed:
@evanchooly , how do i register my own PropertyCodecProvider?
in the Mapper i can register a CodecProvider, but not for properties.
thanks

PR #2172, opened 2022-12-20
Disable or enable validation during update operations

daniel-exceed:
the update did not take into account the update options BypassDocumentValidation, therefore can fail even when we mark to bypass validation

now it will take it into account and disable validation based on the options passed

### DIFF
--- a/core/src/main/java/dev/morphia/query/Modify.java
+++ b/core/src/main/java/dev/morphia/query/Modify.java
@@ -50,6 +50,7 @@ public class Modify<T> extends UpdateBase<T> {
      */
     @Nullable
     public T execute(ModifyOptions options) {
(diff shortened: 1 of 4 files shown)


evanchooly:
There are no tests associated with this change and there are other subclasses of `UpdateBase` that likely could use the same update treatment.

daniel-exceed:
@evanchooly , i added a test case and also added the logic inside Modify class to bypass the validation

evanchooly:
Actually I just noticed this merges the concepts of morphia's model validation and the server's document validation which are, at this moment, two separate concepts.  they should almost certainly be combined but I need to look in to this first.  There's another open issue about the model validation so I'm going to close this in favor of fixing that one.  I will almost certainly come back to this notion as I believe it's the better long term fix.  I just need to think through all the implications and code fixes to make that happen.

PR #2173, opened 2022-12-20
Remove deprecateduse of aggregation

daniel-exceed:
remove use of deprecated aggregation cursor


evanchooly:
I  have this done in a 3.0 branch already.  `master` is still, technically 2.4 and thus would not be eligible for this kind of change due to semver rules.

PR #2190, opened 2023-01-02
change property codec registries from service loader to mapper

daniel-exceed:
fix registration of property codec providers, right now doesnt work, as it uses java service loader that needs module specifications when using java 9+, so i changed the way of adding provider like it's done with regular codec providers, through the mapper options

the bug reproduced:
https://github.com/daniel-exceed/property-codec-service-loader

### DIFF
--- a/core/src/main/java/dev/morphia/mapping/codec/MorphiaMapPropertyCodecProvider.java
+++ b/core/src/main/java/dev/morphia/mapping/codec/MorphiaMapPropertyCodecProvider.java
@@ -21,7 +21,7 @@ import org.bson.codecs.pojo.TypeWithTypeParameters;
 import static dev.morphia.aggregation.codecs.ExpressionHelper.document;
 
 @SuppressWarnings("unchecked")
-class MorphiaMapPropertyCodecProvider extends MorphiaPropertyCodecProvider {
+public class MorphiaMapPropertyCodecProvider extends MorphiaPropertyCodecProvider {
     @Override
     public <T> Codec<T> get(TypeWithTypeParameters<T> type, PropertyCodecRegistry registry) {
         if (Map.class.isAssignableFrom(type.getType())) {
@@ -47,12 +47,12 @@ class MorphiaMapPropertyCodecProvider extends MorphiaPropertyCodecProvider {
         return null;
     }
 
-    private static class MapCodec<K, V> implements Codec<Map<K, V>> {
-        private final Class<Map<K, V>> encoderClass;
-        private final Class<K> keyType;
-        private final Codec<V> codec;
+    public static class MapCodec<K, V> implements Codec<Map<K, V>> {
+        protected final Class<Map<K, V>> encoderClass;
+        protected final Class<K> keyType;
+        protected final Codec<V> codec;
 
-        MapCodec(Class<Map<K, V>> encoderClass, Class<K> keyType, Codec<V> codec) {
+        protected MapCodec(Class<Map<K, V>> encoderClass, Class<K> keyType, Codec<V> codec) {
             this.encoderClass = encoderClass;
             this.keyType = keyType;
             this.codec = codec;
@@ -95,7 +95,7 @@ class MorphiaMapPropertyCodecProvider extends MorphiaPropertyCodecProvider {
             return encoderClass;
         }
 
-        private Map<K, V> getInstance() {
+        protected Map<K, V> getInstance() {
             if (encoderClass.isInterface()) {
                 return new HashMap<>();
             }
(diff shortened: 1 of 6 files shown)


daniel-exceed:
@evanchooly , i opened an issue to document this
https://github.com/MorphiaOrg/morphia/issues/2196

core/src/main/java/dev/morphia/mapping/codec/MorphiaMapPropertyCodecProvider.java
```diff
@@ -21,7 +21,7 @@
 import static dev.morphia.aggregation.codecs.ExpressionHelper.document;
 
 @SuppressWarnings("unchecked")
-class MorphiaMapPropertyCodecProvider extends MorphiaPropertyCodecProvider {
+public class MorphiaMapPropertyCodecProvider extends MorphiaPropertyCodecProvider {
```
evanchooly, line 24: This doesn't need to be made public. your test class need only be a `PropertyCodecProvider`.

core/src/main/java/dev/morphia/mapping/codec/MorphiaMapPropertyCodecProvider.java
```diff
@@ -47,12 +47,12 @@ public <T> Codec<T> get(TypeWithTypeParameters<T> type, PropertyCodecRegistry re
         return null;
     }
 
-    private static class MapCodec<K, V> implements Codec<Map<K, V>> {
-        private final Class<Map<K, V>> encoderClass;
-        private final Class<K> keyType;
-        private final Codec<V> codec;
+    public static class MapCodec<K, V> implements Codec<Map<K, V>> {
```
evanchooly, line 50: same as above.

core/src/test/java/dev/morphia/test/mapping/codec/pojo/PropertyCodecProviderTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+public class PropertyCodecProviderTest extends TestBase {
+    public PropertyCodecProviderTest() {
+        super(MapperOptions.builder()
+                .propertyCodecProvider(new MyCodecProvider())
+                .build());
+    }
+
+    @Test
+    public void testEnumSetPRoperty() {
```
evanchooly, line 35: typo in the method name

core/src/test/java/dev/morphia/test/mapping/codec/pojo/PropertyCodecProviderTest.java
```diff
@@ -0,0 +1,113 @@
+package dev.morphia.test.mapping.codec.pojo;
```
evanchooly, line 1: In general, this test does more than it needs to. It would be much simpler to have your custom provider throw an exception which can be caught to validate that it is properly registered with the `CodecProvider`.  This removes the need to widen the visibility of the internal types as you've done here.

evanchooly (changes requested):

daniel-exceed:
@evanchooly i addressed the points you made. please see my latest commit

core/src/main/java/dev/morphia/mapping/codec/MorphiaMapPropertyCodecProvider.java
daniel-exceed: reverted

core/src/main/java/dev/morphia/mapping/codec/MorphiaMapPropertyCodecProvider.java
daniel-exceed: changed back

core/src/test/java/dev/morphia/test/mapping/codec/pojo/PropertyCodecProviderTest.java
daniel-exceed: fixed

core/src/test/java/dev/morphia/test/mapping/codec/pojo/PropertyCodecProviderTest.java
daniel-exceed: done

evanchooly:
My apologies.  I've been traveling for conferences the last week or so and so I've been a bit delayed.  A suggestion was made in another thread/issue discussion about the use of annotations, like `@MorphiaCodec` or `@MorphiaPropertyCodec` that might be a better/nicer solution than manually registering them and would also work with the module system.  What do you think of that approach?

daniel-exceed:
i think that the annotation solution is more elegant and i support it.
btw, can you say when is the next release planned?

evanchooly:
Closing as there seems to be no issue here. I have updated the docs for the upcoming version to address this topic: https://morphia.dev/morphia/2.4/configuration.html#_some_notes_on_serviceloader

PR #2203, opened 2023-01-12
Flesh out the documentation for the core lifecycle annotations

eheimburg:
No code changes -- just some documentation tweaks. The lifecycle document is a good start, but I still had questions about how each annotation worked, and documenting the annotations seems like the best place to put more info. (It's where I looked first, anyway.)

I tried to only use examples that could be potentially useful -- for example, I didn't add examples of @PostPersist functions that take zero parameters, because I couldn't think of any situation where that would be useful.

This is just a draft -- open to alterations.

### DIFF
--- a/core/src/main/java/dev/morphia/annotations/PostLoad.java
+++ b/core/src/main/java/dev/morphia/annotations/PostLoad.java
@@ -8,7 +8,40 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * Called after the data has been loaded into the java object.
+ * Called after the data has been loaded into the java object. This is a good place to perform 
+ * initialization and sanity-checking of the object.
+ * <p>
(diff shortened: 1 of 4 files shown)


evanchooly:
Thank you for taking the time for the updates.  You'll notice some failures on the CI run.  If you could clean those up, once we get a clean run we can merge.

evanchooly (changes requested):
changes mentioned in the comments

evanchooly:
> Thank you for taking the time for the updates. You'll notice some failures on the CI run. If you could clean those up, once we get a clean run we can merge.

mostly, these change can be made by running `mvn spotless:apply` in `core`.

eheimburg:
also added a missing use-case from @PrePersist. I still didn't mention that it can take a Document parameter -- I can't think of a use-case for it, so treating as an undocumented implementation detail

evanchooly:
Looks good!  Thanks for the patch.

PR #2227, opened 2023-01-20
Fix for the class of an annotated entity being ignored and the annotation improperly applied to other entities.

zenbones:
The code should be self-explanatory. It does fix the problem in my specific instance. I'd be happier being able to run the tests but they are not self-encapsulated and do not run with a 'mvn clean install', and I don't know what the environment requires to successfully run them. In any case, I think you can see the intent of the fix and take it from there.


zenbones:
This also applies a one line fix for pulling custom codec from the MapperOptions as the service loading does not work. I know another contributor has added a similar pull request, but this one is a one line fix that uses arguments already being passed in.

core/src/main/java/dev/morphia/mapping/codec/MorphiaCodecProvider.java
```diff
@@ -4,10 +4,7 @@
 import java.util.HashMap;
```
evanchooly, line 4: The changes here seem unrelated.  At a minimum it looks like the file was reordered so it makes reading the diff complicated.  Can you confirm if these changes are necessary and if not revert.  If they are necessary can you revert the order back to the original order so the important diffs are easier to review?  Thanks.

evanchooly (changes requested):
Thanks for taking the time to fix this.  I have long despised this as it has always seemed overly convoluted.  But it worked.  I'm actually planning on reworking all this with 3.0 (hopefully) but it should still work in the meantime.  In general there are lot of extraneous formatting changes that need to be rolled back so the diff is cleaner.  There should also be tests verifying this fixes the problem included in the PR.

core/src/main/java/dev/morphia/mapping/codec/MorphiaCodecProvider.java
zenbones: Sorry, our in-house formatting standard is different. I can revert all the changes, and put them back in using your formatting. But allow me to explain as well...

EntityModel

The EntityModel stores the lifecycle methods in...

    private final Map<Class<? extends Annotation>, List<ClassMethodPair>> lifecycleMethods;

...which ignores the originating class, i.e. the annotated class. So, If I annotate a class A with a @PrePersist, I get an entry in the map from PrepPersist -> ClassMethodPair holding my pre-persist method. That method is now applied across the instances of class A, and all fields of class A which are also entities. This is at a minimum. Maybe the prepersist method is called on classes other than A, or on fields of A that are not entities, I can't tell. The problem is that the call of the prepersist method of A on anything other than instances of A is both incorrect and yields an error, as other types do have that method. My reproducer shows this problem clearly.

This fix I've produced changes the lifecycle map to...

  private final Map<Class<?>, Map<Class<? extends Annotation>, List<ClassMethodPair>>> lifecycleMethods;

...which is to hold a map **for the annotated class**, which points to a map of lifecycle annotations and methods from that class, and which will be applied only to that class in the later processing. All the other changes to EntityModel os in support of that extended structure. That's it.

Now, the code is not good. Neither the original nor the fix. EntityModel is overly complex and weirdly integrated, But the proposed fix is the least intrusive thing I could think of and seems to work in my case. Again, I'd be happier running the tests, can you explain how to do that?

MorphiaCodecProvider

As noted by others the service loading does not work. Fortunately, the codec can be set on MapperOptions. Maybe that was a response to the failure of service locading? Codecs can also be set on ClientSettings. However, we need codecs set by the user to be checked **first**, and there's no way to know who set what codecs in ClientSettings, so I understand the need for another method so the codecs set with it can take priority. The few lines I inserted into MorphiaCodecProvider...

    CodecProvider codecProvider;

    if ((codecProvider = mapper.getOptions().codecProvider()) != null) {
      // Must go first to user defined codecs take precedent
      propertyCodecProviders.add((PropertyCodecProvider)codecProvider);
    }

...are just grabbing the codec provider from MapperOptions, if it exists, and adding it to the codec providers known to MorphiaCodecProvider, at the head of the list, to be checked first. Again, this code all seems to be awkward. From a user perspective the entry point is not great either, Of course, neither was a service provider. an @CodecProvider annotation would be much better in my opinion. But again, given the code, this seems like the least intrusive fix that worked for me.

zenbones:
Kept original formatting of both EntityModel and MorphiaCodecProvider to make reading and verifying the pull request easier.

core/src/main/java/dev/morphia/mapping/codec/MorphiaCodecProvider.java
```diff
@@ -48,11 +48,19 @@ public MorphiaCodecProvider(Datastore datastore) {
         this.datastore = datastore;
         this.mapper = datastore.getMapper();
 
+        CodecProvider codecProvider;
```
evanchooly, line 51: so this is what I thought.  This change is unrelated to the lifecycle fix and should go in a different PR.  This change can't go in 2.3.1 as it introduces an API change.

evanchooly (changes requested):

evanchooly:
As for writing tests for this, take a look at `dev.morphia.test.mapping.TestLifecycles` for inspiration.  A test in this case might check on the parent and subtype to ensure which method gets called.

core/src/main/java/dev/morphia/mapping/codec/MorphiaCodecProvider.java
zenbones: I understand, but I'll note that without this change I believe the project is broken and so this change should follow soon. Without something like this, custom codecs are ignored for the default codecs. Is it actually an API change to have registered custom codecs actually getting used? That's not so much an API change as a behavioral fix.

zenbones:
It's not about writing tests. I assumed there should already be a test. If not, I guess I could write one as I've already submitted a reproducer. My problem is in running any test. What is the required environmental setup that allows tests to run successfully, because that does not happen locally for me currently?

zenbones:
Thanks, I'll see if I can get tests running, and, if I can, I'll try to add tests.

zenbones:
MorphiaCodecProvider takes a list of PropertyCodecProvider, and as the service provider doesn't work, if providers on MapperOptions are going to be seen by MorphiaCodecProvider, they then need to be a PropertyCodecProvider, but this is not true for the test cases. Looks like this is an internal clash, as MapperOptions takes CodecProvider but only PropertyCodecProvider will make any difference. I can't easily resolve this and all I can do is ignore anything that isn't a PropertyCodecProver in MorphiaCodecProvider, so I'll try that.

All this should probably have had test cases before the new APIs were built out this way, but this stuff would have failed before getting this far.

zenbones:
As far as the EntityModel problem is concerned, I'm trying to make the fix as small as possible. It's hard because EntityModel clearly assumes it can hold data for multiple classes, and that assumption is clearly wrong. The one place that reality is in conflict with the code, that leaks outside of EntityModel itself, is...

```
    public boolean hasLifecycle(Class<? extends Annotation> type) {
        return lifecycleMethods.containsKey(type);
    }
```

This is called from places that do not have a reference to the class they are checking on the existence of the lifecycle method for. And EntityModel tries to hold lifecycle methods for multiple classes. At the very least a parent entity and its attributes, and I'm not sure how far up the chain that goes. The real fix is to refactor EntityModel to hold a single instance in a hierarchy, and to be hierarchy aware with children EntityModel instances for each entity attribute. And then we can have an EntityModels map that holds the relationship of entity class to the EntityModel in the graph, and we can move up or down the graph from there as necessary. But that's a large change to the entire code base.

For me, I think the mistake in understanding comes in with the elimination of @Embedded. That was an important distinction and losing that meta-data leads to the current problem. With @Embdeded it's clear that an EntityModel represents a class annotated with @Entity, and it would need to hold a chain of models for its @Embedded classes only. With the loss of @Embedded, every @Entity needs an EntityModel, and every @Entity can now be used in both ways, as a stand alone class or as part of another model. So we have to look for the existence of @Id to determine meaning, and that annotation can be inherited. Yes, it's there in the hierarchy of inheritance in implementation, but it's no longer clear, to either the model developer using Morphia, or to the Morphia code base itself. And that lack of clarity can be seen in the way EntityModel now squashes annotated methods on its children into itself without regard for their distinctness and independence.

Keeping EntityModel as is I'm going to need to travel the entire call chain of all uses of hasLifecycle() and add the originating entity into the call as a parameter. EntityModel should hold the class it's a model for, but, again, it's no longer a model for a single class, so I think passing that as a parameter is the only option for now.

zenbones:
I am coming to the opinion that MorphiaCodecProvder's handling of custom codecs, and the construction of both EntityModel and PropertyModel are just wrong, and they are strongly tied together, and the combination of these two wrongs just make something more wrong. I have fixes for EntityModel, but MorphiaCodecProvider asks if the EnttyModel has a lifecycle method for the entity class, which is true, so MorphiaCodecProvider sets the LifecycleCodec on the MorphiaCodec, and then the MorphiaCodec proceeds to call the LifecycleCodec on every attribute of the class, and LifeCycle codec assumes each of those values is a document, and when it's not the DocumentWriter throws an exception for being in the wrong state. This project just confuses entities and properties and whose got annotations and who doesn't and what class it's calling lifecycle methods on. I might eventually figure it out, but, if this were a project I inherited, I would rip all of this out and redo it in some way that was clean, and hierarchical and strongly designed.

For us Morphia 1.3.2. works. Unfortunately, it does not work with Mongo 6.0 or the Jakarta namespace. We need something that does. This project is 3 minor versions into 2.x, so I have to assume 2.x was created with very little thought to custom codecs or embedded types. There's some good stuff in it, like Filters, but if custom codecs don't work, and lifecycle methods don't work, it's just not useable. I have to move to Spring Data, which is way overdesigned, and leans heavily on implicit assumptions, and has a terrible class hierarchy, but it will probably work. Or maybe I can grab Morphia 1.3.2 and fork it, replace javax with jakarta, and try swapping out the driver and Bson dependencies and have a forked project that works. I don't love that notion either.

What do you suggest? Is Morphia 2.x going to be clean and complete sometime in the next 2 or 3 months? Can I rip out EntityModel, PropertyModel and MorphiaCodec, do a complete redesign, and have those pull requests merged? I hate to sound like this but my back is against the wall and I'm a bit frustrated to find a project we depend on so heavily in this state.

evanchooly:
The lifecycle stuff is, admittedly, a mess.  It's been hobbled along since well before 1.0 and just hasn't been given as much attention as other parts.  it's on my list to rework in the near future.  I'm not sure what you mean by "MorphiaCodec proceeds to call the LifecycleCodec on every attribute of the class," though.  The `LifecycleCodec` merely wraps the standard `MorphiaCodec` and ensures that the lifecycle methods are called at the appropriate times.

> This project just confuses entities and properties and whose got annotations and who doesn't

It does not.

> and what class it's calling lifecycle methods on. 

This seems a valid bug.

>   I might eventually figure it out, but, if this were a project I inherited, I would rip all of this out and redo it in some way that was clean, and hierarchical and strongly designed.

While I appreciate the feedback, your tone is veering toward the combative here.  I am more than happy to help fix bugs and find resolutions but this kind of "feedback" is not helpful or conducive to a reasonable working environment.  It's very easy to come in to a new codebase and critique decisions that were made years ago without the benefit of the context and drivers behind those decisions.  I'm constantly working on improving and cleaning up the code but apart from occasional contributions from users, it's just me working this in my spare time.  Help is always appreciated.  Disparagement and rudeness are not.

> Is Morphia 2.x going to be clean and complete sometime in the next 2 or 3 months?

Clean is a subjective evaluation so you're free to consider however you'd like.  What "complete" means is completely undefined here but I'm always trying to add new features and improve on old ones.  Always trying to keep up to date with the driver and server features as seem relevant to Morphia.

> Can I rip out EntityModel, PropertyModel and MorphiaCodec, do a complete redesign, and have those pull requests merged?

Almost certainly not as such a major refactoring/redesign mid-version seems ill-advised.  Lifecycles bugs notwithstanding, this code has been used successfully for years and years.  Swooping in and asking to redesign things seems a bit hubristic.  Ultimately, the source of your pain seems centered on the lifecycle code which is fair and i'm more than happy to finally turn my attention to that code as well.  It has gotten a lot of feedback from users to date and so has been a lower priority than other areas.  I'm happy to focus on that for a bit.

> Or maybe I can grab Morphia 1.3.2 and fork it, replace javax with jakarta, 

Apart from the validation bits, I don't know that there is any javax stuff in the mix.  But it has been several years since I've had to dig through the 1.3.2 code.

> and try swapping out the driver and Bson dependencies and have a forked project that works.

That's a tough hill to climb as much of what 1.x depended on has been deprecated/removed from the driver, I believe.

>  I have to assume 2.x was created with very little thought to custom codecs 
Custom codec support is newer.  Custom `PropertyCodec` support has never come up and this is where you're starting to find the rough edges here.

> or embedded types

Embedded types have been supported since long before 1.0.

> It's hard because EntityModel clearly assumes it can hold data for multiple classes, and that assumption is clearly wrong.

`EntityModel` only models information for one type.  You might be talking about how an `EntityModel` might contain `PropertyModel`s for properties from a parent type but that's an internal optimization on how that data is modeled.  But it only concerns itself with that one type.  And each entity type has its own `EntityModel`.

>  I think the mistake in understanding comes in with the elimination of @Embedded. That was an important distinction and losing that meta-data leads to the current problem.

`@Embedded` add no real semantics internally.  It's only real role was to loosen the validation requirement that a mapped entity have an `@Id` field.

> With the loss of @Embedded, every entity needs an EntityModel, 
this was already the case.  all mapped types were backed with an `EntityModel` whether they were annotated with `@Entity` or `@Embedded`.

> and every entity can now be used in both ways, as a stand alone class or as part of another model.

And that's largely a good thing.  In practice, this already happened in code out in the wild.  The distinction between `@Entity` and `@Embedded` grew to be less and less meaningful and was the source of some confusion and support issues.

>  And that lack of clarity can be seen in the way EntityModel now squashes annotated methods on its children into itself without regard for their distinctness and independence.

I don't know what this means.  An `EntityModel` typically does not care about the properties of its children.  One notable difference is in the presence of lifecycle methods and that bookkeeping is limited largely just to note that an entity should be saved/loaded using the `LifecycleCodec`.

All that being said, it's clear that lifecycle support is broken in ways that haven't been caught to date.  I can rework that code as part of a 2.4 release and see if we can't clean that up to everyone's satisfaction.

PR #2233, opened 2023-01-24
BitSet codec

eheimburg:
Adds a codec for `java.util.BitSet` as an array of longs. I've also added support for the legacy driver's save format, as that was very helpful to me when upgrading from Morphia 1.X. This gives users an automatic conversion path if they're still on 1.6 or earlier.


evanchooly:
`mvn test -Dtest=BitSetCodecTest  # -Dmongodb=<some version>`

if you don't have a mongodb running, you can pass that `-D` property to automatically start one.

eheimburg:
For whatever reason, I can't get the `mvn test` syntax to work, but I got all the tests working here, so it's ready for review!

Issue #2340, opened 2023-04-07
Test containers dependency should not be leaked to the library users


cocorossello:
**Describe the bug**
I just noticed that the test container dependency should be in the test scope (SNAPSHOT version).

**To Reproduce**

**Expected behavior**
Dependency should not be leaked

** Please complete the following information: **
 - Server Version:
 - Driver Version:
 - Morphia Version: SNAPSHOT

cocorossello:
https://github.com/MorphiaOrg/morphia/pull/2341

evanchooly:
fixed by PR

PR #2341, opened 2023-04-07
#2340 testcontainers dependency should not be leaked to the library users

cocorossello:


cocorossello:
I see the error, let me try to figure out

evanchooly:
Thanks for the clean up!  :)

PR #2752, opened 2023-12-19
Adding the Maven build cache extension. 

trishagee:
This should speed up builds, particularly when changes are limited to just one module. This extension is open source and free to use.

Currently the Morphia full project build (clean verify) is taking 4-9 minutes locally. Using the build cache:
 - If you make no changes to the code, the project takes 8 seconds to run (locally). I appreciate this is not super useful since one usually wants to make changes! But without the build cache, this would take the full 4-8 minutes
 - If you make a change to Morphia core, the build may take as much time as it usually does, perhaps a little less
 - If you make changes to non-core modules or test classes only, the build may take less time, although I have less experience with how the open source build cache works so this would need some investigation

Feel free to reject this pull request if you think this won't give any benefit or if you want to do some more investigation before applying a new plugin.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,7 @@
+<extensions>
+    <extension>
+        <groupId>org.apache.maven.extensions</groupId>
+        <artifactId>maven-build-cache-extension</artifactId>
+        <version>1.1.0</version>
+    </extension>
+</extensions>


evanchooly:
I love it, actually.  It's been on my list of things to explore that I've just never gotten around to.

evanchooly:
something weird is going on with trying to build locally. it's not looking at the generated folders for sources so I've commented it out for now until I have a chance to dig in to the why.  it's weird that it builds on github...

trishagee:
Oh weird. I don't know enough about how this specific plugin works, but I know the Develocity one I work with does sometimes need tweaking for generated sources. I'll try and do some research.

Issue #2828, opened 2024-01-24
UpdateOperations removeAll Duplicates Key


danielmerchant:
**Describe the bug**

 I believe I've found a bug in the UpdateOperation#removeAll method - wondering if this is a known issue. I'm using version 2.4.9 right now.

I have an array on my document that stores embedded documents:

```
@Entity("Documents")
public class MyDocument {

  private List<EmbeddedDocument> embeddedDocs;
}

@Entity
public class EmbeddedDocument {
  private String field1;
  private String field2;
  private String field3;
}
```

I've added two embedded docs to one MyDocument:

```
EmbeddedDocument doc1 = new EmbeddedDocument("foo1", "bar1", "baz1");
EmbeddedDocument doc2 = new EmbeddedDocument("foo2", "bar2", "baz2");

MyDocument root = new MyDocument(List.of(doc1, doc2));

datastore.save(root);
```

I've added the document to my DB, so there should be one document in Documents that looks like this:

```
mongos> db.Documents.find({})
{
  "embeddedDocs":[
    {
      "field1": "foo1",
      "field2": "bar1",
      "field3": "baz1"
    },
    {
      "field1": "foo2",
      "field2": "bar2",
      "field3": "baz3"
    }
  ]
} 
```

Now, I want to remove entities stored in the embeddedDocs array that have the value field2 = "bar1". In mongosh form:
```
mongos> db.Documents.updateMany({}, {"$pull": {"embeddedDocs": {"field2": "bar1"}}})
```
And in Java:
```
Query<MyDocument> query = Datastore.createQuery(MyDocument.class);

EmbeddedDocument onlyField2 = new EmbeddedDocument(null, "bar1", null);

UpdateOperations<MyDocument> removeAllOp = 
datastore.createUpdateOperations(MyDocument.class)
  .removeAll("embeddedDocs", onlyField2);

datastore.update(query, removeAllOp);
```

In Morphia 1, it works. In Morphia 2, when stepping through in the debugger, I found that instead of generating the expected update document:
```
{"$pull": {"embeddedDocs": {"field2": "bar1"}}}}
```

We get a duplicated key, which leads to a document  like this:
```
{"$pull": {"embeddedDocs": {"embeddedDocs": {"field2": "bar1"}}}}
```
Which doesn't remove the entity from the array.

**To Reproduce**
1. Create a document with an embedded array of documents.
2. Attempt to remove one document from the embedded array using a condition via the `#removeAll` method.
[...]

PR #2834, opened 2024-02-01
Fix for UpdateOperations removeAll Duplicates Key #2828

dfjones:
Implements a fix for #2828 by removing an implicit use of `Filters.eq` for the value passed in to the `pull` operator. 

This PR adds test coverage for the affected code. Here, the value can be either an `Object` or an instance of `Filter`. The implementation of the `PullOperator` has been updated to handle both.


dfjones:
I think I found a bug in the implementation here...working on updating the test cases and revising.

dfjones:
Just pushed my revising implementation. I think this is solid and more similar to how other update operators are implemented.

dfjones:
Pushed a change to add test coverage for the rest of the legacy update API here. Good news, they all pass. 😄

core/src/main/java/dev/morphia/query/updates/UpdateOperators.java
```diff
@@ -244,13 +243,13 @@ public static PopOperator pop(String field) {
     /**
      * The $pull operator removes from an existing array all instances of a value or values that match a specified condition.
      *
-     * @param field  the field to update
-     * @param filter the filter to apply
+     * @param field the field to update
+     * @param value the instance of a value or the filter to apply
```
evanchooly, line 247: "the value or filter to use" is a little less of a mouthful. :)

core/src/main/java/dev/morphia/query/updates/UpdateOperators.java
```diff
@@ -244,13 +243,13 @@ public static PopOperator pop(String field) {
...
      * @return the update operator
      * @update.operator $pull
      */
-    public static UpdateOperator pull(String field, Filter filter) {
-        return new PullOperator(field, filter);
+    public static UpdateOperator pull(String field, Object value) {
```
evanchooly, line 251: technically this break bincompat which I've tried to avoid but I don't think it'll effectively matter and the tools I had in place to check this kind of thing are broken anyway. but something to keep an eye on if users break strangely,.

core/src/test/java/dev/morphia/test/TestLegacyUpdate.java
```diff
@@ -0,0 +1,428 @@
+package dev.morphia.test;
```
evanchooly, line 1: I'm guessing this was just copied over from the "modern" query tests and adjust to use the legacy query code?

core/src/main/java/dev/morphia/query/updates/UpdateOperators.java
dfjones: Yea, I thought about that while working on it. If there's a better way to achieve this without breaking compatibility, I'd happily switch to it. 

I thought about adding a new method here, something like `UpdateOperator pullValue(String field, Object value)`. Let me know if something like that is a better approach. Also, is there any concern about changes to `PullOperator`? I see that's marked internal, but not sure what kind of standards there are around that.

core/src/test/java/dev/morphia/test/TestLegacyUpdate.java
dfjones: No, these are net new, since I had already written this to cover `removeAll` and test for the bug in #2828. But I can take a look at the modern update tests and see if they make sense to port over.

dfjones: It looks like it'd be more comprehensive to port over `TestUpdateOperations` (perhaps with the addition of test coverage specific to #2828) over the tests I added here. Let me know if you'd like me to do that as part of this PR.

evanchooly: i think this is probably fine, honestly. a lot of this has changed in master so it has a limited shelf life anyway.

core/src/main/java/dev/morphia/query/updates/UpdateOperators.java
evanchooly: I mean, we might leave the old `Filter` method have it call the new `Object` method but I'm not sure it's worth it just to satisfy the pedantry of a check that's no longer actively enforced. This is technically a bug fix to the API so if someone complains we can have that conversation then.

dfjones: Makes sense and sounds good to me. Thanks!

PR #3037, opened 2024-05-15
Overridable method for creating morphia codec provider

bendherville:
Allow subclasses to provide a custom morphia codec provider


evanchooly:
@dependabot recreate

evanchooly:
Thanks for the PR but `Datastore` isn't really meant to be extended and those kinds of configuration overrides are being migrate to `MorphiaConfig`.

PR #3038, opened 2024-05-17
Prefer to load custom property codec providers before default codec providers.

danielmerchant:
Like the PR title says - prefer to load user-defined custom property codec providers before the library default codec providers. This allows users to override default behavior to provide customizable codec logic for `Map` and `Collection` types, as well as their subclasses.

Added a test here to validate that the NoOp provider I added would load correctly, but I had to add it in a package separate from the existing tests to get access to the protected methods. Not sure if that's something I should avoid here, but generally it's considered good practice to have unit tests live in the same package as the source it's testing. I also had to add a `META-INF/services/MorphiaPropertyCodecProvider` file here to add the No-Op codec provider so that the service loader could access it. I tried monkeying around with ways to get it to load for _only_ the test I wrote, but no avail.

### DIFF
--- a/core/src/test/java/dev/morphia/mapping/codec/MorphiaCodecProviderTest.java
+++ b/core/src/test/java/dev/morphia/mapping/codec/MorphiaCodecProviderTest.java
@@ -1,0 +1,29 @@
+package dev.morphia.mapping.codec;
+
+import java.util.List;
+
+import dev.morphia.Datastore;
+import dev.morphia.test.TestBase;
+
+import org.bson.codecs.pojo.PropertyCodecProvider;
+import org.testng.annotations.Test;
+
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertTrue;
+
+public class MorphiaCodecProviderTest extends TestBase {
+
+    @Test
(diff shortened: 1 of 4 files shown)


core/src/test/java/dev/morphia/mapping/codec/MorphiaCodecProviderTest.java
```diff
@@ -0,0 +1,29 @@
...
+import org.bson.codecs.pojo.PropertyCodecProvider;
+import org.testng.annotations.Test;
+
+import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.assertTrue;
+
+public class MorphiaCodecProviderTest extends TestBase {
+
+    @Test
+    public void NewMorphiaCodecProvider_NoOpPropertyCodecProviderInConfig_ExpectedOrderOfProvidersReturned() {
```
evanchooly, line 17: this is quite the mouthful.  maybe a simpler name?  maybe i'll just rename after the merge.

danielmerchant: Fine by me. Sometimes I go a little overboard with test naming.

PR #3068, opened 2024-06-03
Run tests in parallel on different JVM forks

bdemers:
Low hanging fruit for running tests in parallel is forking multiple JVMs (`1C`, one per core) and having each one run tests sequentially.
This cuts the build time from 5m to 2m on my laptop.

~~NOTE: I was writing up my thoughts on other ways to run the tests in parallel without needing to fork, but then I realized all the tests are using `"mongodb://localhost:27017/" + TEST_DB_NAME;`, so I'm not sure how running my tests worked?~~
Answer: https://github.com/MorphiaOrg/morphia/blob/29a32b8969aa8bc915b4cf3a0b257840965dc083/core/src/test/java/dev/morphia/test/MorphiaTestSetup.java#L62-L65

@evanchooly I was chatting with @trishagee to get the tests to run in parallel.


evanchooly:
I think what you're running into is that `MorphiaContainer` was added to try and solve a particular problem ("clear" boundaries around configuration reuse and avoiding remapping classes) but it doesn't really solve that problem and makes things way more complicated than it's worth. I've half intended to remove it but never really had a big enough motivator but I think maybe now is a good time.

If each class runs off in its own forked env/thread/whatever, we can simply rebuild it all each time (yes, we'll remap code more than we need but i'm working on optimizing that, too) and simply run each test with its own database.  Then we won't have to worry about dropping databases while trying to write to them.  In fact, there's no need to drop the databases anymore since they all run on ephemeral testcontainers instances.  Let me see about unwinding the `MorphiaContainer` junk and see if that makes your life any easier.

PR #3492, opened 2025-06-18
Fix incorrectly returned value in MorphiaMapCodec.java

pputerla:
We've noticed a bug in `MorphiaMapCodec.java` that is it does not return the map it creates but rather always `Map.of()`

This PR fixes that. Test included.

From the user's perspective it's a regression when switching from 2.4.x to 2.5.x

But is especially visible when using `aggregate` and mapping into `Map.class` type.

### DIFF
--- a/core/src/main/java/dev/morphia/mapping/codec/MorphiaMapCodec.java
+++ b/core/src/main/java/dev/morphia/mapping/codec/MorphiaMapCodec.java
@@ -70,7 +70,7 @@ public class MorphiaMapCodec implements Codec<Map> {
         }
 
         reader.readEndDocument();
-        return Map.of();
+        return map;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


evanchooly:
1.  That's a horrifying bug.
2. I wonder how/why the tests never caught that!

pputerla:
> ah, because `dev.morphia.mapping.codec.MorphiaMapPropertyCodecProvider.MapCodec` is what does the decoding and `MorphiaMapCodec` isn't involved. This codec is mostly used to serialize "internal" types to the driver (options class, driver types, etc.) related to the functioning of morphia itself rather than user types. So that's why it never showed in testing. May I ask how you encountered it? It might help to highlight other potential usage patterns that aren't properly tested.

oh, we just used aggregate query. Usually we map the output to target pojo, but this time it was explicitly mapped to generic Map.class, as shown in the test:
```
getDs().aggregate(SomeClass.class).execute(Map.class)
```

PR #4159, opened 2026-02-12
Added class loader option to the project

ThomasWega:
Found this issue out when creating a Paper (Minecraft)  plugin. Minecraft plugins each have their own ClassLoader. This makes it impossible for me to map classes as the class loader morphia is using is just a different one. 

There was a workaround in earlier version as I was able to find here
https://stackoverflow.com/a/63698526
Even a suggested feature here: https://github.com/MorphiaOrg/morphia/pull/1394

If I'm not mistaken, that's no longer the case. This pretty recent issue on github was a similar thing, but it suggested to use Thread.currentThread().getContextClassLoader() which 2.5.2 does I think, however that just does not solve my issue as you can see from this screenshot.
image
The classloader is just different.

I've tried adding the option to input custom classloader in the project. Take this more as a draft than a finished thing. I'm looking for feedback and will make any change you request. There is also still one place that uses the old classloader which I've marked with FIXME. I'm quite honestly not sure if all places require to use the custom class loader.

I have tested this version on my server with my project and it seems to work completely fine. It fixed the previous issues where no classes could be mapped and when they were mapped manually they would not resolve when being taken from the database.

I also would've done this change for 3.0.0 but I just wasn't able to get it to work with my maven. It's using the maven 4 which is in preview and seems like my PC is having issues loading it even though I've upgraded.

I'm also available on the discord server as `ThomasWega` if you want to discuss anything there

### DIFF
--- a/core/src/main/java/dev/morphia/mapping/codec/references/ReferenceCodec.java
+++ b/core/src/main/java/dev/morphia/mapping/codec/references/ReferenceCodec.java
@@ -164,7 +164,7 @@ public class ReferenceCodec extends BaseReferenceCodec<Object> implements Proper
                 try {
                     id = datastore.getCodecRegistry()
                             .get(datastore.getMapper().getClass(document))
-                            .decode(new DocumentReader(document, datastore.getClassLoader()), decoderContext);
+                            .decode(new DocumentReader(document, datastore.getMapper().getClassLoader()), decoderContext);
                 } catch (CodecConfigurationException e) {
                     throw new MappingException(Sofia.cannotFindTypeInDocument(), e);
                 }
@@ -176,7 +176,7 @@ public class ReferenceCodec extends BaseReferenceCodec<Object> implements Proper
             if (refId instanceof Document) {
                 refId = datastore.getCodecRegistry()
                         .get(Object.class)
-                        .decode(new DocumentReader((Document) refId, datastore.getClassLoader()), decoderContext);
+                        .decode(new DocumentReader((Document) refId, datastore.getMapper().getClassLoader()), decoderContext);
             }
             id = new DBRef(ref.getDatabaseName(), ref.getCollectionName(), refId);
         }
@@ -374,13 +374,14 @@ public class ReferenceCodec extends BaseReferenceCodec<Object> implements Proper
         Codec<?> codec = getDatastore().getCodecRegistry().get(getEntityModelForField().getType());
         return value.stream()
                 .filter(v -> v instanceof Document && ((Document) v).containsKey("_id"))
-                .map(d -> codec.decode(new DocumentReader((Document) d, datastore.getClassLoader()), DecoderContext.builder().build()))
+                .map(d -> codec.decode(new DocumentReader((Document) d, datastore.getMapper().getClassLoader()),
+                        DecoderContext.builder().build()))
                 .collect(Collectors.toList());
     }
 
     MorphiaReference<?> readDocument(Document value) {
         final Object id = getDatastore().getCodecRegistry().get(Object.class)
-                .decode(new DocumentReader(value, datastore.getClassLoader()), DecoderContext.builder().build());
+                .decode(new DocumentReader(value, datastore.getMapper().getClassLoader()), DecoderContext.builder().build());
         return readSingle(id);
     }
 


evanchooly (changes requested):
Thanks for the PR and the detailed write-up — the motivation is solid. Environments like Minecraft Paper plugins (and OSGi, etc.) where the context classloader doesn't match the application classloader are a real pain point, and Morphia should support custom classloaders.

That said, I found several issues with the current implementation that would need to be addressed before this can be merged.

---

## Critical Issues

### 1. `MorphiaConfigHelper` reflection bug — will crash at runtime

The change in `MorphiaConfigHelper.getEntry()`:

```java
.getDeclaredConstructor(morphiaConfig.getClass())
.newInstance(morphiaConfig);
```

`morphiaConfig.getClass()` returns the **runtime class** (e.g., `ManualMorphiaConfig`), but all the converters declare their constructor parameter as `MorphiaConfig` (the interface). So `getDeclaredConstructor(ManualMorphiaConfig.class)` will throw `NoSuchMethodException`. This should be `getDeclaredConstructor(MorphiaConfig.class)`.

### 2. SmallRye config loading is broken

The converters (`ClassNameConverter`, `DiscriminatorFunctionConverter`, `NamingStrategyConverter`, etc.) are referenced via `@WithConverter` annotations and instantiated by SmallRye Config using **no-arg constructors**. This PR removes all no-arg constructors by adding a required `MorphiaConfig` parameter. SmallRye won't be able to instantiate them, which would break `MorphiaConfig.load()` entirely.

### 3. Circular dependency in config loading

`MorphiaConfig.load(path, classLoader)` needs to create a config, but the converters used during loading now require a `MorphiaConfig` instance that doesn't exist yet — chicken-and-egg problem.

### 4. `classLoader()` on the `MorphiaConfig` interface is incompatible with SmallRye

`MorphiaConfig` is a `@ConfigMapping` interface. Adding `ClassLoader classLoader()` without `@WithDefault` means SmallRye would try to map it from the config file, but `ClassLoader` isn't a serializable config value. This needs to be excluded from SmallRye mapping or handled differently.

---

## Design Issues

### 5. `getClassLoader()` on public `Datastore` interface

This is added without `@MorphiaInternal`, making it part of the public API contract. This is an implementation detail that shouldn't be exposed publicly — users shouldn't need to call `datastore.getClassLoader()`.

### 6. `DiscriminatorLookup` holds a full `Mapper` reference

This introduces a bidirectional dependency: `Mapper` → `DiscriminatorLookup` → `Mapper`. It would be cleaner to pass just the `ClassLoader` or a `Supplier<ClassLoader>` rather than the entire `Mapper`.

### 7. Incomplete implementation (the FIXME)

As you noted, `Conversions.java` still uses the old classloader:

```java
// FIXME still old classloader!
return Thread.currentThread().getContextClassLoader().loadClass(className);
```

### 8. `ManualMorphiaConfig.classLoader()` default is evaluated lazily each call

```java
return orDefault(classLoader, Thread.currentThread().getContextClassLoader());
```

Each call could return a **different** classloader depending on which thread calls it. The default should be captured once at construction time.

---

## Suggested Approach

Rather than threading `MorphiaConfig` through SmallRye converters (which breaks config loading), consider setting the classloader on the config **after** loading and using it only at runtime — not during config parsing. The converters don't need a custom classloader during config load since Morphia's own classes are being resolved at that point. The custom classloader is primarily needed at runtime for discriminator lookup, entity scanning, reference resolution, and proxy creation.

evanchooly:
sounds good.  those comments were claude's (i'm not that verbose) but i was working on my own response as well.  i'll wait for the follow up changes, though.  :)

ThomasWega:
I suppose ready for review now. I've tried to keep it as clean and simple. Everything that's not super internal should be backwards compatible. Just added new methods with new parameters and kept the old ones. 

All usages of `Thread.currentThread().getContextClassLoader()` have now been replaced, except if you are using the default (current) approach in which case it defaults to that. 
Also converters use the contextClassLoader as it was said above those don't need custom class loader.

Let me know if there's anything more you need me to do :)

core/src/main/java/dev/morphia/aggregation/AggregationImpl.java
```diff
@@ -354,7 +354,7 @@ public Aggregation<T> addStage(Stage stage) {
         return this;
     }
 
-    private static class MappingCursor<R> implements MongoCursor<R> {
+    private class MappingCursor<R> implements MongoCursor<R> {
```
evanchooly: Turns out this class isn't even used.  I'll delete it on the 2.5 branch, though, so don't worry about it just now.

core/src/main/java/dev/morphia/config/MorphiaConfig.java
```diff
@@ -53,6 +53,16 @@ static MorphiaConfig load() {
         return load(MorphiaConfigHelper.MORPHIA_CONFIG_PROPERTIES);
     }
 
+    /**
+     * Tries to load a configuration from the default location using the given classloader.
+     *
+     * @param classLoader the classloader to use when loading resources from the classpath
+     * @return the loaded config
+     */
+    static MorphiaConfig load(ClassLoader classLoader) {
```
evanchooly: Adding new methods to an API in a patch release is verboten by semver. I'm trying to decide how much semver matters to me with the 2.5.  On the one hand, semver is important.  On the other hand, 3.0 is much slower in coming than I'd hoped (though I'm actually closer than I thought to what might be considered a beta or RC if I downscope a little and push things to a 3.1...) and adhering to such a rule seems a little counterproductive in that light.  Maybe make a 2.6 out of all this and satisfy both?  That gets close to too many active branches, though, so maybe mark 2.4 and 2.5 as "done" and any future fixes go in to 2.6 alone?  Or maybe just 2.4 and 2.6 since I think 2.5 is mildly breaking because of the driver upgrade and some internal incompatibilities...  Thoughts?

core/src/main/java/dev/morphia/mapping/codec/ArrayFieldAccessor.java
```diff
@@ -84,6 +86,6 @@ private Object convert(Object o, Class<?> type) {
 
             return newArray;
         }
-        return Conversions.convert(o, type);
+        return Conversions.convert(o, type, classLoader);
```
evanchooly: Maybe a clean approach would be to make `Conversions.convert()` not static and create an instance of it from the classloader configured on `MorphiaConfig`.  then the convert method signature doesn't really change and the interesting parts of the API that are nominally available for non-Morphia usage stays smaller (a partial focus of 3.0 is paring down the API surface area.  so many overloads....)  Then we just pass in the `Conversions` instance (stored on the `Datastore`?) when we create these kinds of classes.  It reduce the amount of classloader parameters being passed from class to class, i think.

core/src/main/java/dev/morphia/Datastore.java
```diff
@@ -635,6 +635,11 @@ default <T> UpdateResult update(Query<T> query, dev.morphia.query.UpdateOperatio
     @MorphiaInternal
     Mapper getMapper();
 
+    @MorphiaInternal
+    default ClassLoader getClassLoader() {
```
evanchooly: nein!  😁

core/src/main/java/dev/morphia/DatastoreImpl.java
```diff
@@ -641,7 +645,7 @@ public <T> void refresh(T entity) {
                 .iterator()
                 .next();
 
-        refreshCodec.decode(new DocumentReader(id), DecoderContext.builder().checkedDiscriminator(true).build());
+        refreshCodec.decode(new DocumentReader(id, getClassLoader()), DecoderContext.builder().checkedDiscriminator(true).build());
```
evanchooly: this argument would be the `Conversions` instance field we'd add to the `DatastoreImpl`, e.g.

core/src/main/java/dev/morphia/Morphia.java
```diff
@@ -89,4 +89,16 @@ public static Datastore createDatastore(MongoClient mongoClient, MorphiaConfig c
...
+    /**
+     * Creates a Datastore configured via config file
+     *
+     * @param mongoClient the client to use
+     * @param config      the configuration to use
+     * @param classLoader the classloader to use when scanning for entities and codecs. If null, the default classloader will be used.
+     * @return a Datastore that you can use to interact with MongoDB
+     * @since 3.0.0
+     */
+    public static Datastore createDatastore(MongoClient mongoClient, MorphiaConfig config, ClassLoader classLoader) {
```
evanchooly: This wouldn't be necessary.  We'd just use `config.classLoader()` inside `DatastoreImpl` to get what we need.

evanchooly (changes requested):
Overall, I like the direction.  There have been a few "using the wrong classloader" bugs "recently."  They're hard to suss out because they require operating in contexts I don't have access to (glassfish/wildfly apps, minecraft plugins, etc.) and this change would help close some of those gaps I think we've gotten lucky so far to have not run in to.  There are some simplifications in the comments that I think might serve to shrink this PR down even further.  if I get time before you get to them, i'll see about pushing a patch to this PR showing what i'm thinking.  But if you beat me to it, I won't complain either.  Thanks for all the work you've put in so far.  This revision is definitely cleaner than the first one.

core/src/main/java/dev/morphia/config/MorphiaConfig.java
ThomasWega: I personally don't use this method and for anyone who's having issue with classloaders he can get around using config files as well just by using the ManualMorphiaConfig. 
I'm inclining more towards 2.6 release myself, however I don't know how big the previous releases were to jump up a minor release version, but reading the definition, I think this PR suits perfectly for 2.6 (MINOR release). Depends whether you call this a new functionality or a bug fix I guess :D

```
MAJOR version when you make incompatible API changes
MINOR version when you add functionality in a backward compatible manner
PATCH version when you make backward compatible bug fixes
```

core/src/main/java/dev/morphia/config/MorphiaConfig.java
```diff
@@ -62,7 +72,20 @@ static MorphiaConfig load() {
...
+     * Parses and loads the configuration found at the given location
+     *
+     * @param path        the location of the configuration to load. This can be a file path, a classpath resource, a URL, etc.
+     * @param classLoader the classloader to use when loading resources from the classpath
+     *
+     * @return the loaded configuration
+     * @since 3.0
+     */
+    static MorphiaConfig load(String path, ClassLoader classLoader) {
+        List<ConfigSource> configSources = classPathSources(path, classLoader);
```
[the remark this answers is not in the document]
ThomasWega: You are saying we should input classloader once the morphia config is constructed, but I've added the classloader here so it would even find the config class in the first place. Cuz in my specific environment if I went ahead and used a config file, I'd need to input the classloader just for morphia to be able to even find the file inside the plugin's resources

Wouldn't your proposal break this if we need to first initialize the morphia config and then set the classloader? How will morphia know which class loader to use to look for the file?

core/src/main/java/dev/morphia/mapping/codec/pojo/MorphiaCodec.java
```diff
@@ -91,7 +91,7 @@ public Object generateIdIfAbsentFromDocument(Object entity) {
         if (!documentHasId(entity)) {
             if (idProperty != null) {
                 if (ObjectId.class.equals(idProperty.getType()) || String.class.equals(idProperty.getType())) {
-                    idProperty.setValue(entity, convert(new ObjectId(), idProperty.getType()));
+                    idProperty.setValue(entity, convert(new ObjectId(), idProperty.getType(), datastore.getClassLoader()));
```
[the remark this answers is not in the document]
ThomasWega: The datastore only uses the mapper getter. I guess I could remove it from the datastore and always get mapper and then the classloader from there, but I don't see a reason to do that when I can have this "shortcut"

evanchooly:
um.  [this PR](https://github.com/MorphiaOrg/morphia/pull/4174) was supposed to be based on your branch but claude messed up.  i can give it another go or if you like the changes, doing a merge in to your branch would be trivial as well.  apologies for the delay.

ThomasWega:
Hey! Any news? No hurry though

ThomasWega:
Hey. Apologies for the delay and thank you, appreciate it. Is there anything more you need me to do? I will test the new release in the upcoming days and update you on that

PR #4257, opened 2026-05-19
feat: reference deduplication and cycle-safe decoding via DecodeSession

evanchooly:
Add a per-document ThreadLocal decode session that caches entity instances
by (collection, id). Nested activate() calls reuse the existing session so
all @Reference fetches within one document decode share the same cache.

- DecodeSession: lightweight ThreadLocal cache with activate/deactivate;
  activate() returns true only for the root caller (owns deactivation)
- MorphiaCursor: activates a session around next()/tryNext(); only
  deactivates if it was the root activator
- EntityDecoder: pre-registers the entity instance via peekId() (mark/reset)
  before decodeProperties(), enabling cycle detection
- ReferenceCodec: checks session cache before firing DB queries for both
  single and collection references
- TestReferences: adds testReferenceDeduplication (within-document) and
  testCyclicReferenceDoesNotStackOverflow

https://claude.ai/code/session_01NRjHAe69rLVqEZYZHnBsBF


evanchooly:
### Code review

Found 2 issues:

1. `ConstructorCreator`-backed entities are pre-registered with a null/zero-valued instance. `instanceCreator.getInstance()` is called before `decodeProperties()`, which for all-args-constructor entities (Java records, Kotlin data classes) constructs the instance immediately with all parameters at their zero defaults. Because `ConstructorCreator` caches the first-constructed instance, subsequent `set()` calls during `decodeProperties()` fall into the `instance != null` branch and attempt to set fields via reflection — which silently fails for final fields. The session cache then holds a permanently broken shell that is returned to any caller who looks it up.

https://github.com/MorphiaOrg/morphia/blob/a64cc50463c7b2f78fc2a4f5d065eaf02e9d7718/core/src/main/java/dev/morphia/mapping/codec/pojo/EntityDecoder.java#L47-L62

2. `LifecycleDecoder.decode()` overrides `EntityDecoder.decode()` without calling `super.decode()`, so all session registration logic (`peekId`, `session.register`) is entirely skipped for entities annotated with `@PreLoad`/`@PostLoad`. Cyclic or duplicate `@Reference` graphs involving lifecycle-annotated entities will still cause stack overflows or duplicate DB fetches, silently defeating the feature for that class of entities.

https://github.com/MorphiaOrg/morphia/blob/a64cc50463c7b2f78fc2a4f5d065eaf02e9d7718/core/src/main/java/dev/morphia/mapping/codec/pojo/LifecycleDecoder.java#L35-L62

